PR #535, opened 2014-05-10
Tests, that page HTML isn’t rebuilt by `getContent` method

aik099:
Original PR (to `develop` branch): #462


jcalderonzumba:
@aik099 this will break tests for https://github.com/jcalderonzumba/MinkPhantomJSDriver, see the problem is that PhantomJS **does** changes original HTML  returned from the server because phantomjs does not have a method to get "raw" HTML send by the server.

stof:
Closing this, as that's not possible to implement for Selenium-based drivers either.

PR #596, opened 2014-08-19
add moveTo and click without a element

gimler:
Adds the    possibility to work without elements just with coordinates.
- Add  `moveTo`
- Add x and y coordinates to `mouseOver`
- Change click events to work without elements


drzraf:
AFAICT without `moveTo($element, $pixels)` it's impossible to simulate a slider dragged in a direction.

PR #616, opened 2014-10-28
Add empty check for path in WebAssert cleanUrl()

jonesio:
PHP parse_url does not always return a 'path' component which causes a PHP strict error as `cleanUrl` references the array key directly in the return value. This checks for a path and defaults to root ('/') if empty.


stof:
OK, I'm merging this and adding the test myself

stof:
I just tagged the 1.6.1 release with this fix, instead of waiting for the 1.7 release.

PR #627, opened 2014-12-16
fixing issue when url is /

gouaille:


stof:
I merged #616 instead as it was first and I preferred the way the code was formatted.

PR #633, opened 2015-01-08
Implement the conversion of XPath to NodeElement in CoreDriver

stof:
The goal is to ease the development of Mink 2.0 by allowing drivers to be compatible with both Mink 1.7 and 2.0 on a single code base. The only change required for drivers is now handled at the level of the CoreDriver shipped in Mink when drivers should to override findElements rather than find (see https://github.com/minkphp/MinkZombieDriver/pull/58/files for what is required currently).
This will allow decoupling the 2.0 version of Mink from a new major version of drivers (the new major version of ZombieDriver should rather be related to the full rewrite suggested in https://github.com/minkphp/MinkZombieDriver/issues/104 if it ever happens than to the Mink refactoring for instance)

### DIFF
--- a/src/Driver/CoreDriver.php
+++ b/src/Driver/CoreDriver.php
@@ -10,6 +10,7 @@
 
 namespace Behat\Mink\Driver;
 
+use Behat\Mink\Element\NodeElement;
 use Behat\Mink\Exception\UnsupportedDriverActionException;
 use Behat\Mink\Session;
 
@@ -22,11 +23,16 @@ use Behat\Mink\Session;
 abstract class CoreDriver implements DriverInterface
 {
     /**
+     * @var Session
+     */
+    private $session;
+
+    /**
      * {@inheritdoc}
      */
     public function setSession(Session $session)
     {
-        throw new UnsupportedDriverActionException('Setting the session is not supported by %s', $this);
+        $this->session = $session;
     }
 
     /**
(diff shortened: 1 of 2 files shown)


aik099:
OK.

stof:
this is how it will impact drivers regarding to the Mink versions:

- current stable release: ``~1.6`` => Mink 1.6 and 1.7 (I don't plan a 1.8 release, but it would also support it if any)
- next driver release switching to implementing ``findElements`` instead: ``~1.7|~2.0``, supporting both the latest 1.x versions (so people can use it easily) and the next Mink version (for people dealing with the necessary upgrades for BC breaks in the API). any work in drivers will then support both.

src/Behat/Mink/Driver/CoreDriver.php
```diff
@@ -90,6 +96,28 @@ public function getContent()
...
+     *
+     * @see find()
+     *
+     * @param string $xpath
+     *
+     * @return string[] The XPath of the matched elements
+     *
+     * @throws UnsupportedDriverActionException When operation not supported by the driver
+     */
+    protected function findElements($xpath)
```
stof: I'm not 100% satisfied about the name of this method (it does not return Element objects after all). But I don't have a better idea. What about you @aik099 ?

aik099: So it finds xpaths of elements within element with given xpath. We don't have much words to operate with here. Only `find`, `element`, `xpath`, but since `xpath` is implementation detail we can't use it method name.

Maybe we can call it `findElementSelectors`.

stof: I also see an issue with this name: we also use the concept of Selector in Mink, and it is not exactly the same think than here (even though it is related)

stof: @everzet do you have any suggestion about the name ?

@aik099 what do you think ?

everzet: `matchElements()` ?

stof: Well, my issue with the ``findElements`` name is the ``elements`` part, because it does not return elements objects. So ``matchElements`` is not better in this regard

aik099: The method accepts xpaths and returns xpaths and it's unlikely, that this behavior will change any time soon. Maybe we should expose `xpath` word in method name.

Another set of possible names:
* `findElementXpaths`
* `findElementLocators`
* `findSubXpaths`
* `findLocatorsFromXpath`

stof: I will go for ``findElementXpaths``

stof:
@aik099 are we good now ?

aik099:
Yes.

stof:
All driver PRs are now done

PR #623, opened 2014-11-30
Add support for header assertions

erikwebb:
Depends on https://github.com/minkphp/MinkBrowserKitDriver/pull/90

PR #636, opened 2015-01-23
Add support for header assertions

stof:
This continues the work of #623. It removes the changes in the driver API as we already have a method to get the response headers. This avoids forcing drivers to duplicate the header retrieval, and it let us implement the case insensitive lookup in a single place (to deal with header names properly).
It also adds the missing tests for the new method in the Session class.

### DIFF
--- a/src/Driver/CoreDriver.php
+++ b/src/Driver/CoreDriver.php
@@ -256,14 +256,6 @@ abstract class CoreDriver implements DriverInterface
     /**
      * {@inheritdoc}
      */
-    public function getResponseHeader($name)
-    {
-        throw new UnsupportedDriverActionException('Response headers are not available from %s', $this);
-    }
-
-    /**
-     * {@inheritdoc}
-     */
     public function getResponseHeaders()
     {
         throw new UnsupportedDriverActionException('Response headers are not available from %s', $this);
--- a/src/Driver/CoreDriver.php
+++ b/src/Driver/CoreDriver.php
@@ -256,6 +256,14 @@ abstract class CoreDriver implements DriverInterface
     /**
      * {@inheritdoc}
      */
+    public function getResponseHeader($name)
+    {
+        throw new UnsupportedDriverActionException('Response headers are not available from %s', $this);
+    }
+
+    /**
+     * {@inheritdoc}
+     */
     public function getResponseHeaders()
     {
         throw new UnsupportedDriverActionException('Response headers are not available from %s', $this);
--- a/src/Driver/DriverInterface.php
+++ b/src/Driver/DriverInterface.php
@@ -182,18 +182,6 @@ interface DriverInterface
     public function setRequestHeader($name, $value);
 
     /**
-     * Returns specific response header.
-     *
-     * @param string $name
-     *
-     * @return string|null
-     *
-     * @throws UnsupportedDriverActionException When operation not supported by the driver
-     * @throws DriverException                  When the operation cannot be done
-     */
-    public function getResponseHeader($name);
-
-    /**
      * Returns last response headers.
      *
      * @return array
--- a/src/Driver/DriverInterface.php
+++ b/src/Driver/DriverInterface.php
@@ -182,6 +182,18 @@ interface DriverInterface
     public function setRequestHeader($name, $value);
 
     /**
+     * Returns specific response header.
+     *
+     * @param string $name
+     *
+     * @return string|null
+     *
+     * @throws UnsupportedDriverActionException When operation not supported by the driver
+     * @throws DriverException                  When the operation cannot be done
+     */
+    public function getResponseHeader($name);
+
+    /**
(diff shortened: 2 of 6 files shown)


stof:
I'm not sure we actually need the regex-based assertions for header values. What do you think @aik099 ?

src/Behat/Mink/WebAssert.php
```diff
@@ -146,6 +146,106 @@ public function statusCodeNotEquals($code)
...
+     *
+     * @param string $name
+     * @param string $value
+     *
+     * @throws ExpectationException
+     */
+    public function responseHeaderNotContains($name, $value)
+    {
+        $actual = $this->session->getResponseHeader($name);
+        $actual = preg_replace('/\s+/u', ' ', $actual);
```
stof: note to myself: if we keep these assertions, this replacement should be removed. It does not make sense in headers.

src/Behat/Mink/WebAssert.php
```diff
@@ -146,6 +146,106 @@ public function statusCodeNotEquals($code)
...
+     * @throws ExpectationException
+     */
+    public function responseHeaderContains($name, $value)
+    {
+        $actual = $this->session->getResponseHeader($name);
+        $actual = preg_replace('/\s+/u', ' ', $actual);
+        $regex  = '/'.preg_quote($value, '/').'/ui';
+        $message = sprintf('The text "%s" was not found anywhere in the "%s" response header.', $value, $name);
+
+        $this->assert((bool) preg_match($regex, $actual), $message);
```
aik099: Jusing using `strpos` will be better.

src/Behat/Mink/WebAssert.php
```diff
@@ -146,6 +146,106 @@ public function statusCodeNotEquals($code)
...
+     * @throws ExpectationException
+     */
+    public function responseHeaderNotContains($name, $value)
+    {
+        $actual = $this->session->getResponseHeader($name);
+        $actual = preg_replace('/\s+/u', ' ', $actual);
+        $regex  = '/'.preg_quote($value, '/').'/ui';
+        $message = sprintf('The text "%s" was found in the "%s" response header, but it should not.', $value, $name);
+
+        $this->assert(!preg_match($regex, $actual), $message);
```
aik099: Jusing using `strpos` will be better.

aik099:
> I'm not sure we actually need the regex-based assertions for header values.

Yes, they can be replaced with `strpos` usage. But partial matching against header value is a needed check IMO because in complex header values you may want to assert presence of small part (e.g. attachment name in long header value).

stof:
@aik099 updated

src/Behat/Mink/WebAssert.php
```diff
@@ -146,6 +146,102 @@ public function statusCodeNotEquals($code)
...
+     * @param string $regex
+     *
+     * @throws ExpectationException
+     */
+    public function responseHeaderMatches($name, $regex)
+    {
+        $actual = $this->session->getResponseHeader($name);
+        $message = sprintf('The pattern "%s" was not found anywhere in the "%s" response header.', $regex, $name);
+
+        $this->assert((bool) preg_match($regex, $actual), $message);
```
aik099: The `preg_match` can return 3 values:
* `1` in case of found
* `0` in case of not found
* `false` in case of error

I think we should explicitly compare with `1` to be on safe side: `1 === preg_match....`. Inverted IF should be used in `responseHeaderNotMatches` method as well.

src/Behat/Mink/WebAssert.php
```diff
@@ -146,6 +146,102 @@ public function statusCodeNotEquals($code)
...
+     *
+     * @param string $name
+     * @param string $regex
+     *
+     * @throws ExpectationException
+     */
+    public function responseHeaderNotMatches($name, $regex)
+    {
+        $actual = $this->session->getResponseHeader($name);
+        $message = sprintf('The pattern "%s" was found in the text of the "%s" response header, but it should not.', $regex, $name);
```
aik099: This line is too long. Please split across several lines.

mink_webassert_responseheadernotmatches_toolong


Maybe we should:
1. inline `$message` variable
2. from result extract only 1st argument of `sprintf` to a `$message` variable

src/Behat/Mink/WebAssert.php
stof: Well, a match is 1, and it is the only case which can become true when casting to boolean. It is fine IMO.

aik099: 1 isn't match count :smile:

aik099: As long as it works we can keep it.

stof:
@aik099 are we good now ?

aik099:
Yes. And also yes for that other PR in MinkExtension that will use new methods.

PR #637, opened 2015-01-24
Increase the code coverage (line coverage is now 100%)

stof:
We had a few non-executed statements in our testsuite.


stof:
Note that I'm aware that 100% line coverage is not the silver bullet at all. The good news is coming in the next PR :smile:

aik099:
OK.

PR #638, opened 2015-01-24
Add the humbug configuration

stof:
This adds the configuration file to run mutation testing with Humbug. It allows to know whether our testsuite actually covers our library.
Currently, Humbug applyes 119 mutations and reports 19 of them as being escaped (i.e. not catched by tests). However this is wrong. Among these 19, 12 should be catched but there is a bug in Humbug: https://github.com/padraic/humbug/issues/71

I intentionally don't configure Travis to run them. The Humbug output needs to be analyzed by a developer to know whether the escaped mutations are relevant or no (some mutations are harmless, which will trigger false positives). Thus, Humbug does not even return a failure exit code when there are some escaped mutations (probably because of the need of a human review btw), which would make it a pain to integrate on CI).
This config file is here to make it easy to run Humbug ourselves from time to time.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -3,3 +3,4 @@
 phpunit.xml
 composer.lock
 vendor
+/build
--- a/humbug.json
+++ b/humbug.json
@@ -1,0 +1,12 @@
+{
+    "timeout": 10,
+    "source": {
+        "directories": [
+            "src"
+        ]
+    },
+    "logs": {
+        "text": "build/humbuglog.txt",
+        "json": "build/humbuglog.json"
+    }
+}


aik099:
I've also executed Humbug but on QA-Tools some time ago. I still need to analyze it's findings to see if some tests doesn't all of it.

stof:
Do you agree with including the config file in the repo directly ?

aik099:
I'm not sure. The mutation testing isn't done on regular basis and test suite don't change radically too often. That suggests, that having Humbug config in repo isn't really a thing that every contributor on a project needs to have.

But on the other hand this pretty small config, so you can commit it anyways :)

aik099:
And current config doesn't apply to driver test suite.

stof:
> And current config doesn't apply to driver test suite.

This is logical. The driver testsuite does not cover the Mink code. It is here to ensure consistency between driver implementations.

stof:
With the fixed versions of humbug, we are now catching 112 mutations on 119 (94%, which is pretty good). Among the 7 remaining ones, 4 of them are harmless and 2 are about cosmetic in exception messages. Our testsuite coverage is quite good.

aik099:
So no changes are required to catch more zombies?

stof:
@aik099 well, we have 1 mutation worth catching, and 2 which might be catched (but not sure it is actually worth it for them). The 4 harmless ones either cannot be catched (changing the sleep time from 10000µs to 10001µs in wait loops) or are not worth it (ensuring that the exception code is actually 0 when calling the parent exception constructor, as we don't actually care about the code)

stof:
I ran it again. We are now at 86% because of the code added in #641 where the testsuite does not cover all cases for the detection of escaped Xpath in the NamedSelector

stof:
Closing this, as we still haven't merged that config. If we want to revisit that, we should use Infection, which replaced humbug

PR #640, opened 2015-01-28
Added test for dumping with empty tag

WouterJ:
See https://github.com/minkphp/MinkBrowserKitDriver/pull/93 for the problem

### DIFF
--- a/driver-testsuite/tests/Basic/ContentTest.php
+++ b/driver-testsuite/tests/Basic/ContentTest.php
@@ -21,6 +21,18 @@ class ContentTest extends TestCase
         );
     }
 
+    public function testDumpingEmptyElements()
+    {
+        $this->getSession()->visit($this->pathTo('/index.html'));
+
+        $element = $this->getAssertSession()->elementExists('css', '#empty');
+
+        $this->assertEquals(
+            "An empty <em></em> tag should be rendered with both open and close tags.",
+            $element->getHtml()
+        );
+    }
+
     /**
      * @dataProvider getAttributeDataProvider
      */
--- a/driver-testsuite/tests/Basic/ContentTest.php
+++ b/driver-testsuite/tests/Basic/ContentTest.php
@@ -29,7 +29,7 @@ class ContentTest extends TestCase
 
         $this->assertEquals(
             "An empty <em></em> tag should be rendered with both open and close tags.",
-            $element->getHtml()
+            trim($element->getHtml())
         );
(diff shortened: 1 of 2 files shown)


aik099:
How much drivers pass this new test? I think drivers that use `.innerHTML` for returning page content, as Sahi, might not pass this test.

driver-testsuite/tests/Basic/ContentTest.php
```diff
@@ -21,6 +21,18 @@ public function testOuterHtml()
...
     }
 
+    public function testDumpingEmptyElements()
+    {
+        $this->getSession()->visit($this->pathTo('/index.html'));
+
+        $element = $this->getAssertSession()->elementExists('css', '#empty');
+
+        $this->assertEquals(
+            "An empty <em></em> tag should be rendered with both open and close tags.",
```
stof: I think this might fail if some drivers preserve the leading and trailing spaces in the HTML, as you have some of them in the HTML

WouterJ: add a trim around the result

stof:
@WouterJ can you answer the question asked by @aik099 ?

WouterJ:
I have no idea how other drivers will support this. However, I don't understand how `.innerHTML` would break things.

aik099:
> I have no idea how other drivers will support this.

You just need to run this test on all of the drivers to know.

WouterJ:
You can remove "just", as that would mean installing sahi and zombie, downloading all drivers and installing all dependencies for all drivers... Can't this just be merged, followed by a test rebuild of each driver to see which drivers fail and then fix it? If this test fails, it was a bug in the driver anyway. This PR doesn't make drivers more unstable.

aik099:
>  as that would mean installing sahi and zombie, downloading all drivers and installing all dependencies for all drivers..

Sure. Me & @stof have the setup to run tests on all drivers locally. It's recommended for any contributor.

stof:
I'm currently running this against many drivers. I will post the result

stof:
BrowserKitDriver: failing (the PR referenced above is fixing it)
GoutteDriver: inherited from BrowserKitDriver
Selenium2Driver: passing
ZombieDriver: passing
SeleniumDriver: passing
SahiDriver: passing

WouterJ:
Thank you @stof! I just rebased the BrowserKitDriver PR

PR #641, opened 2015-02-04
Moved the escaping of the xpath locator to the NamedSelector

### LINKED ISSUE #384: NamedSelector::translateToXPath error on 'content' selector
Hi there!

I'm new in Behat so it might be something I'm doing wrong.
So doing:
```php
$selector = new \Behat\Mink\Selector\NamedSelector();
$xpath = $selector->translateToXPath( array( 'content', "a string here" ) );
```
will lead to:
```
 Unknown named selector provided: "./descendant-or-self::*[contains(normalize-space(.), omgomgogm)]". Expected one of (fieldset, field, link, button, link_or_button, content, select, checkbox, radio, file, optgroup, option, table)
```

Also doing:
```php
$page->find( 'named', array( 'content', "a string here" ) );
```
will lead to:
```
      error:_sahi.setServerVarPlain('___lastValue___52499ba54203f', (function(){
          var count = 0;
          while (_byXPath("(//html/./descendant-or-self::*[contains(normalize-space(.), a string here)])["+(count+1)+"]")) count++;
          return count;
      })())
      SyntaxError: The expression is not a legal expression.
      No trace available
      <a href='/_s_/dyn/Log_getBrowserScript?href=null&n=1601'><b>Click for browser script</b></a>

```

Thanks

stof:
The NamedSelector has no valid reason to expect receiving an escaped value for the XPath locator. Calling code has no reason to know that the locator will be inserted in an XPath query.
The NamedSelector still support passing an escaped value for BC reasons but this is deprecated.
The method SelectorsHandler::xpathLiteral has been deprecated as well.
Fixes #384

This allows fixing the issue in Mink 1.7 rather than only 2.0, which allows people to migrate their code progressively (which is better)

### DIFF
--- a/src/Element/Element.php
+++ b/src/Element/Element.php
@@ -85,6 +85,8 @@ abstract class Element implements ElementInterface
      * Returns selectors handler.
      *
      * @return SelectorsHandler
+     *
+     * @deprecated Accessing the selectors handler in the element is deprecated as of 1.7 and will be impossible in 2.0.
      */
     protected function getSelectorsHandler()
     {
@@ -156,7 +158,7 @@ abstract class Element implements ElementInterface
             return $items;
         }
 
-        $xpath = $this->getSelectorsHandler()->selectorToXpath($selector, $locator);
+        $xpath = $this->selectorsHandler->selectorToXpath($selector, $locator);
         $xpath = $this->xpathManipulator->prepend($xpath, $this->getXpath());
 
         return $this->getDriver()->find($xpath);
--- a/src/Selector/NamedSelector.php
+++ b/src/Selector/NamedSelector.php
@@ -10,6 +10,8 @@
 
 namespace Behat\Mink\Selector;
 
+use Behat\Mink\Selector\Xpath\Escaper;
+
 /**
  * Named selectors engine. Uses registered XPath selectors to create new expressions.
  *
@@ -157,12 +159,15 @@ XPATH
 .//*[%idOrNameMatch%]
 XPATH
     );
+    private $xpathEscaper;
 
     /**
      * Creates selector instance.
      */
     public function __construct()
     {
+        $this->xpathEscaper = new Escaper();
+
         foreach ($this->replacements as $from => $to) {
             $this->replacements[$from] = strtr($to, $this->replacements);
         }
@@ -217,7 +222,7 @@ XPATH
         $xpath = $this->selectors[$selector];
 
         if (null !== $locator) {
-            $xpath = strtr($xpath, array('%locator%' => $locator));
+            $xpath = strtr($xpath, array('%locator%' => $this->escapeLocator($locator)));
         }
 
         return $xpath;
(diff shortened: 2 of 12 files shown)


aik099:
It looks like some kind of code variation from 2.0 branch, but don't understand how it correlates with other commits from #384 task, that were made in 2013 year.

stof:
@aik099 it is indeed a variation of the work done in https://github.com/stof/Mink/commit/ebf26f8522590d0ed05598315ce888df37e54b11 for #542 but keeping a BC layer instead of forcing the upgrade. The 2.0 PR will then be updated to include only the removal of the BC layer.

aik099:
I like the direction we're going with the drivers to allow safe transition to Mink 2.0.

stof:
this one is not about drivers (#633 is enough for that) but about userland code (I think that after this change, the upgrade should be painless except for some very custom use cases)

stof:
@aik099 is it OK for you ?

src/Element/Element.php
```diff
@@ -82,16 +82,6 @@ protected function getDriver()
     }
 
     /**
-     * Returns selectors handler.
-     *
-     * @return SelectorsHandler
-     */
-    protected function getSelectorsHandler()
```
aik099: This will surely break something within QA-Tools, once merged.

aik099: I guess I need to do `Escaper::escapeLiteral` when I need to escape xpath instead of doing 

```php
$selectors_handler = $this->getSelectorsHandler();
$selectors_handler->xpathLiteral($text)
```

Was this `Escaper` class part of Mink 1.6 release?

src/Selector/NamedSelector.php
```diff
@@ -235,4 +240,24 @@ protected function registerReplacement($from, $to)
     {
         $this->replacements[$from] = $to;
     }
+
+    private function escapeLocator($locator)
+    {
+        // If the locator looks like an escaped one, don't escape it again for BC reasons.
+        if (
+            preg_match('/^\'[^\']*+\'$/', $locator)
```
aik099: It's inverted version of `Escaper::escapeLiteral` I guess.

src/Element/Element.php
aik099: Also what could bullet-proof way of detecting if used Mink does automatic locator escaping by NamedSelector or I need to manually escape it upfront? I'm asking because of https://github.com/qa-tools/qa-tools/blob/develop/library/QATools/QATools/HtmlElements/Element/Form.php#L86 construct.

src/Selector/NamedSelector.php
stof: It detects strings looking like the output of the Xpath escaping, yes

src/Element/Element.php
stof: Once Mink 1.7 is released, the easiest way would probably be to bump the dependency to Mink 1.7 (which autoescapes)

stof: and yes, I should not remove this method for BC. It will be removed only in 2.0

stof: @aik099 you could use ``method_exists(NamedSelector::class, 'escapeLocator')`` to detect that this PR is included in the codebase if you want to keep the Mink 1.6 compatibility.

And when building your own Xpath, you should indeed use the Escaper class (which is available in 1.6.0)

stof:
@aik099 anything else to do before the merge ?

aik099:
Nope. You can merge.

PR #643, opened 2015-02-21
Remove duplicated DOM selection in js_test.html

### LINKED ISSUE #532: Reduce JS selector duplication in "js_test.php"
Right now same jQuery selector is queried several times (in same function) which is forces jQuery to search for element again and again.

We can speed up a process by doing some refactorings on it (as PhpStorm inspection suggests).

stof:
Closes #532


aik099:
OK.

PR #648, opened 2015-05-07
Updated copyright to 2015

maqnouch:


aik099:
That's why I don't add year to any of my licenses :)

maqnouch:
@aik099 You should add the current year to the list of copyright years if and only if you make a copyrightable change to the file. Trivial patches like adding null pointer checks and the like are generally considered to not be copyrightable. Many projects have an explicit policy of requiring an assignment of copyright for patches over a certain length threshold (generally around 20 lines) because it is assumed that anything that long probably represents at least a slight amount of creative thought. You should also never remove any listed copyright years.
:D

PR #650, opened 2015-05-13
[POC] add a way to change the xpath prefix

### LINKED ISSUE #649: WebAssert (and getPage) should be able to dive into xml documents too
As (not really) discussed in #408, it should be possible to query something else than `//html`-prefixed xpath queries.

I have a working diff here and wanted to know if you'd be interested in looking at it.

Cheers!

docteurklein:
Fixes #649.
needs tests. Thoughts @stof ?

Usage:

       $mink->assertSession(null, '//opendata')->elementExists('xpath', $selector);

My first impression is that I don't like to mess with optional `$xpath` arguments :|
Also, I changed the signature of some methods.

### DIFF
--- a/src/Element/DocumentElement.php
+++ b/src/Element/DocumentElement.php
@@ -10,6 +10,8 @@
 
 namespace Behat\Mink\Element;
 
+use Behat\Mink\Session;
+
 /**
  * Document element.
  *
@@ -17,6 +19,21 @@ namespace Behat\Mink\Element;
  */
 class DocumentElement extends TraversableElement
 {
+    private $xpath;
+
+    /**
+     * Initializes node element.
+     *
+     * @param string  $xpath   element xpath
+     * @param Session $session session instance
+     */
+    public function __construct($xpath, Session $session)
+    {
+        $this->xpath = $xpath;
+
+        parent::__construct($session);
+    }
+
     /**
      * Returns XPath for handled element.
      *
@@ -24,7 +41,7 @@ class DocumentElement extends TraversableElement
      */
     public function getXpath()
     {
-        return '//html';
+        return $this->xpath;
     }
 
     /**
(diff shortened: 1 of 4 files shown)


aik099:
Maybe we should approach this by specifying custom DocumentElement instance instead of passing `//html` all over the place?

Currently there is single `DocumentElement` instance per session and you can't change it over time to query XML and then HTML documents.

docteurklein:
I thought about a `setPage` on the session object, but it doesn't sounds right.
But passing `//html` all over the place doesn't sound better :)

One could then do: 

    $session->setPage(new DocumentElement('//root', $session));

and basta.

src/Element/DocumentElement.php
```diff
@@ -17,6 +19,21 @@
...
 {
+    private $xpath;
+
+    /**
+     * Initializes node element.
+     *
+     * @param string  $xpath   element xpath
+     * @param Session $session session instance
+     */
+    public function __construct($xpath, Session $session)
```
stof: this signature change is a BC break.

src/Session.php
```diff
@@ -118,9 +117,9 @@ public function getDriver()
      *
      * @return DocumentElement
      */
-    public function getPage()
+    public function getPage($xpath = '//html')
     {
-        return $this->page;
+        return $this->page = $this->page ?: new DocumentElement($xpath, $this);
```
stof: this logic is broken in case the page already exists and you pass a different XPath

stof:
I would vote against this as it seems out of the scope of Mink and creates a bunch of issues

src/Element/DocumentElement.php
stof: see the testsuite btw

docteurklein:
@aik099 @stof seems legit. Do you have an idea how I could achieve the same goal without this?

aik099:
Closing, because associated issue was closed. See associated issue discussion for an answer.

PR #655, opened 2015-06-19
Fixed WebAssert::cleanUrl assuming front controller

Chekote:
We're using Behat to test a PHP application that is not currently using any framework. We have a simple test where we use a button to navigate from one page to another. e.g. from /page1.php to /page2.php

The button is currently broken, and does not actually leave /page1.php, but when we use `Then I should be on "/page2.php"`  the test passes even though we are still on /page1.php.

We have narrowed this down to the cleanUrl() calls in WebAssert::addressEquals and WebAssert::getCurrentUrlPath, which seems to be stripping out the php script. The only reason I can think of for Mink to do this is that it's assuming the script is a front controller.

Please let me know if I'm not understanding this correctly. If I am, I believe this pull request should resolve the issue.

Thanks.

### DIFF
--- a/src/WebAssert.php
+++ b/src/WebAssert.php
@@ -750,7 +750,7 @@ class WebAssert
         $fragment = empty($parts['fragment']) ? '' : '#'.$parts['fragment'];
         $path = empty($parts['path']) ? '/' : $parts['path'];
 
-        return preg_replace('/^\/[^\.\/]+\.php/', '', $path).$fragment;
+        return preg_replace('/^\/[^\.\/]+\.php\//', '/', $path).$fragment;
     }
 
     /**
--- a/tests/WebAssertTest.php
+++ b/tests/WebAssertTest.php
@@ -52,6 +52,23 @@ class WebAssertTest extends \PHPUnit_Framework_TestCase
         $this->assertCorrectAssertion('addressEquals', array('/'));
     }
 
+    public function testAddressEqualsEndingInScript()
+    {
+        $this->session
+            ->expects($this->exactly(2))
+            ->method('getCurrentUrl')
+            ->will($this->returnValue('http://example.com/script.php'))
+        ;
+
+        $this->assertCorrectAssertion('addressEquals', array('/script.php'));
+        $this->assertWrongAssertion(
+            'addressEquals',
+            array('/'),
+            'Behat\\Mink\\Exception\\ExpectationException',
+            'Current page is "/script.php", but "/" expected.'
+        );
+    }
+
     public function testAddressNotEquals()
     {
         $this->session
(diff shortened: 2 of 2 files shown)


Chekote:
This is in response to [my question](https://groups.google.com/forum/#!topic/behat/lAPHaUXHRM0) posted to the Behat discussion group.

aik099:
I have no idea, what purpose of the code was, but it was added at same time the `WebAssert` class was created in 233c6fcc0dbe7114ae039cf85ece6691cf73aee7 commit. Back then the `cleanUrl` method was called `cleanScriptnameFromPath`.

Obviously it prevents websites that use url schemes like below from working:

* `/one-page.php` and `/another-page.php`
* `/index.php?page=one-page` and * `/index.php?page=another-page`

Chekote:
Yup. The pull request should fix your second example also.

Chekote:
@stof Good Morning Sir! Could you please let me know what I need to do to get this pull request merged?

Thanks ^_^

aik099:
Maybe we can introduce url modes, so that user can choose which mode the `cleanUrl` works:

* keep PHP script in url
* keep query parameters in url
* both
* none - as it currently works

Chekote:
Are you sure that there's a use case where you would need to keep the script if it did not have a trailing slash? I can't think of any myself.

If it's a front controller, then it should always have a trailing slash, which will still function correctly given my change.

aik099:
> Are you sure that there's a use case where you would need to keep the script if it did not have a trailing slash?

Yes, this `/index.php?page=one-page` case exactly.

Chekote:
You *can* remove the php script in that instance.

But there shouldn't be any need to. If you are testing that the url `/?page=one-page` works, that's what you should be using in your steps.

If I assert that I should be on `/index.php?page=one-page` and `/?page=one-page` matches (or visa versa). I would consider that a bug.

Chekote:
Sorry, my question was worded wrong. What I meant was:

> Are you sure that there's a use case where you would need to keep the script if it __DOES__ have a trailing slash? I can't think of any myself.

Without this pull being merged, your example will possibly fail (depending on the configuration of the web server). If this is merged, it will work.

aik099:
Without this PR merged these 2 urls are considered equal, but they are not:

* `/script.php/something/else`
* `/another_script.php/something/else`

I have no idea why @everzet removed PHP script filename from url initially.

Chekote:
Those would still be considered equal, even after the pull request is merged. I was trying to avoid changing that specific behavior. There is a legitimate need for that because if you're using Symfony for example, your base url in your behat.yml would be:

`http://website.dev/app_test.php`

But your Behat step would likely say something like:

```cucumber
I should be on "/my/page"
```

Which would result in the full url of:

`http://website.dev/app_test.php/my/page`

In this scenario, you would want `/my/page` and `/app_test.php/my/page` to be considered equal.

aik099:
Ah, then the need to remove PHP file from url was driven by way how Symfony builds urls. Good to know.

Chekote:
Is there anything else I need to do before this can be merged? We can't move forward with out tests until this is resolved. Thanks! ^_^

Chekote:
@stof Good Morning Sir! Could you please let me know what I need to do to get this pull request merged?

Chekote:
Does anyone know what a guy has to do to get a pull request merged around here? >.>

aik099:
Technically I can, but I waited for @stof 2nd opinion on this after he is back from his vacation. Apparently he was overwhelmed with GitHub notifications and haven't seen one about this PR yet.

Chekote:
Okey dokey. Thanks for the update!

aik099:
@stof , ping.

stof:
Looks good to me

aik099:
Merging. Thanks, @Chekote .

Chekote:
Woot! \o/

PR #658, opened 2015-06-29
Add placeholder to the elementNotFound exception

jordi12100:
It was unclear for me that you can find an element by placeholder.

### DIFF
--- a/src/Element/TraversableElement.php
+++ b/src/Element/TraversableElement.php
@@ -109,7 +109,7 @@ abstract class TraversableElement extends Element
         $button = $this->findButton($locator);
 
         if (null === $button) {
-            throw new ElementNotFoundException($this->getDriver(), 'button', 'id|name|title|alt|value', $locator);
+            throw new ElementNotFoundException($this->getDriver(), 'button', 'id|name|title|alt|value|placeholder', $locator);
         }
 
         $button->press();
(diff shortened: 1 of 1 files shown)


aik099:
Please double check that `placeholder` attribute even can be used with elements, that searched using https://github.com/minkphp/Mink/blob/master/src/Selector/NamedSelector.php#L22-L161 , because currently you're adding `|placeholder` for button finding as well, but they surely don't have placeholders.

src/Element/TraversableElement.php
```diff
@@ -109,7 +109,7 @@ public function pressButton($locator)
         $button = $this->findButton($locator);
 
         if (null === $button) {
-            throw $this->elementNotFound('button', 'id|name|title|alt|value', $locator);
+            throw $this->elementNotFound('button', 'id|name|title|alt|value|placeholder', $locator);
```
stof: this does not make sense. Buttons don't have a placeholder

aik099: I said the same in my comment.

jordi12100:
@aik099 is right:
"The placeholder attribute works with the following input types: text, search, url, tel, email, and password".

I've updated the PR. Only added it to the fillField method.

aik099:
Also maybe it's worth to send PR to documentation (https://github.com/minkphp/docs) as well if we're mentioning `id|name|label|value` in there currently.

jordi12100:
Just checked. It is already mentioned there.
**field**
Searches for a form field by its id, name, label or placeholder.

stof:
@jordi12100 could you rebase your PR to fix conflicts ?

jordi12100:
@stof, done! And the checks passed :wink:

stof:
thanks

PR #659, opened 2015-07-20
add ability to re-run replace

fugi:
Usefull especialy when register new replacement or selector.

### DIFF
--- a/src/Selector/NamedSelector.php
+++ b/src/Selector/NamedSelector.php
@@ -168,14 +168,27 @@ XPATH
     {
         $this->xpathEscaper = new Escaper();
 
-        foreach ($this->replacements as $from => $to) {
-            $this->replacements[$from] = strtr($to, $this->replacements);
-        }
-
-        foreach ($this->selectors as $alias => $selector) {
-            $this->selectors[$alias] = strtr($selector, $this->replacements);
-        }
+        $this->replace();
     }
+
+    /**
+     * Replace placeholders with replacements
+     *
+     * @return NamedSelector
+     */
+    public function replace()
+    {
+    	foreach ($this->replacements as $from => $to) {
+    		$this->replacements[$from] = strtr($to, $this->replacements);
+    	}
+
+    	foreach ($this->selectors as $alias => $selector) {
+    		$this->selectors[$alias] = strtr($selector, $this->replacements);
+    	}
+
+    	return $this;
+    }
+
 
     /**
      * Registers new XPath selector with specified name.


aik099:
Please explain use case of proposed code change.

You should be registering new selectors in constructor prior to parent constructor call instead.

fugi:
How can I owerride default class when it is constructed in _SelectorsHandler_ constructor explicite?
https://github.com/minkphp/Mink/blob/master/src/Selector/SelectorsHandler.php#L34

Furthermore I think that this is better than extending parent class. For example I add new selector and also I want to use another developer selector. There is no way to combine this two class besides copy&paste.

aik099:
> How can I owerride default class when it is constructed in SelectorsHandler constructor explicite?
> https://github.com/minkphp/Mink/blob/master/src/Selector/SelectorsHandler.php#L34
> Furthermore I think that this is better than extending parent class. For example I add new selector and also I want to use another developer selector.

You can register your own selectors handler (that might extend default one if you wish) under same name as any of build-in selectors handlers. Any use case besides just extending?

I suppose default behavior of doing partial match first and exact match send isn't what you need. In that case you need to use `named_exact` instead of `named` when you're looking for elements like this:

```php
// for exact match only
$registerForm = $page->find('named_exact', array('field', 'field-name-here');

// for partial match only
$registerForm = $page->find('named_partial', array('field', 'field-name-here');

// for partial and then exactly if partial didn't find anything match only
$registerForm = $page->find('named', array('field', 'field-name-here');
```

fugi:
OK, but this is hell. For example I have two new selectors(MySelector, AnotherSelector) so I need to make 2 new classes(MyNamedSelector, AnotherNamedSelector). When I want tu use exact and partial matches I need 4 classes(MyNamedSelectorExact, MyNamedSelectorPartial, AnotherNamedSelectorExact, AnotherNamedSelectorPartial). And when I want to use it i need to do it like this:

```php
// for my selector exact match
$registerForm = $page->find('my_named_selector_exact', array('my_selector', 'field-name-here');

// for my selector partial match
$registerForm = $page->find('my_named_selector_partial', array('my_selector', 'field-name-here');

// for another selector exact match
$registerForm = $page->find('another_named_selector_exact', array('another_selector', 'field-name-here');

// for another selector partial match
$registerForm = $page->find('another_named_selector_partial', array('another_selector', 'field-name-here');
```

aik099:
Sorry for confusion. Let's start over if you don't mind.

Please explain problem you're trying to solve (use case) forgetting for a moment about this PR and `replace` method it provides. Maybe I can recommend you another solution that doesn't involve code changing.

aik099:
Sorry for the delay @fugi .

If I get you right you want to add new `selector` to both `PartialNamedSelector` and `ExactNamedSelector` and gets `replacements` replaced inside of it.

That is partially possible with this code:

```php
$selectorsHandler = new \Behat\Mink\Selector\SelectorsHandler();

$namedPartialSelector = $selectorsHandler->getSelector('named_partial');
$namedPartialSelector->registerNamedXpath('new_name', 'new_xpath');

$namedExactSelector = $selectorsHandler->getSelector('named_exact');
$namedExactSelector->registerNamedXpath('new_name', 'new_xpath');

$session = new \Behat\Mink\Session($driver, $selectorsHandler);
```

The only issue with above approach (that doesn't involve extending classes) is fact, that `registerNamedXpath` doesn't replace replacements inside user provided xpath.

If that is what you're after, then please update PR to fix behavior of `registerNamedXpath` method then (+tests).

PR #662, opened 2015-08-29
Silence deprecations by default

stof:
This follows the Symfony convention to make our deprecation warning less invasive (PHPUnit and Behat are both triggering exceptions on deprecation errors by default, thus breaking the BC layers).
Such deprecations warnings can be logged when using the Symfony error handler or the Symfony PHPUnit bridge (or other tools implementing the same convention).

There is no such tool for Behat yet, but this is on my TODO list.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -19,6 +19,10 @@
         "symfony/css-selector":  "~2.0"
     },
 
+    "require-dev": {
+        "symfony/phpunit-bridge": "~2.7"
+    },
+
     "suggest": {
         "behat/mink-browserkit-driver": "extremely fast headless driver for Symfony\\Kernel-based apps (Sf2, Silex)",
         "behat/mink-goutte-driver":     "fast headless driver for any app without JS emulation",
--- a/tests/Selector/NamedSelectorTest.php
+++ b/tests/Selector/NamedSelectorTest.php
@@ -54,14 +54,13 @@ abstract class NamedSelectorTest extends \PHPUnit_Framework_TestCase
 
     /**
      * @dataProvider getSelectorTests
+     * @group legacy
      */
     public function testEscapedSelectors($fixtureFile, $selector, $locator, $expectedExactCount, $expectedPartialCount = null)
     {
         // Escape the locator as Mink 1.x expects the caller of the NamedSelector to handle it
         $escaper = new Escaper();
         $locator = $escaper->escapeLiteral($locator);
-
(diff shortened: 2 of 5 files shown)


composer.json
```diff
@@ -19,6 +19,10 @@
         "symfony/css-selector":  "~2.0"
     },
 
+    "require-dev": {
+        "symfony/phpunit-bridge": "~2.7"
```
aik099: What does this do?

tests/Selector/NamedSelectorTest.php
```diff
@@ -54,6 +54,7 @@ public function testSelectors($fixtureFile, $selector, $locator, $expectedExactC
 
     /**
      * @dataProvider getSelectorTests
+     * @group legacy
```
aik099: I wonder if this group is somehow related to added `@` in other places.

tests/Selector/NamedSelectorTest.php
```diff
@@ -61,8 +62,6 @@ public function testEscapedSelectors($fixtureFile, $selector, $locator, $expecte
         $escaper = new Escaper();
         $locator = $escaper->escapeLiteral($locator);
 
-        $this->iniSet('error_reporting', -1 & ~E_USER_DEPRECATED);
```
aik099: This is done somewhere else I presume.

composer.json
aik099: I've read http://symfony.com/blog/new-in-symfony-2-7-phpunit-bridge and now I get the clear picture.

stof: it make test fail if you call a deprecated code path in a non-legacy test

tests/Selector/NamedSelectorTest.php
stof: this tells the PHPUnit bridge that it is OK to use deprecated APIs in this test, because we are precisely testing the legacy API to ensure that we don't break the BC layer.

tests/Selector/NamedSelectorTest.php
stof: this is not needed anymore:

- the deprecation is silenced and so not triggered as exception
- when using the PHPunit bridge, it relies on the legacy group to detect that it is OK to call deprecated APIs

composer.json
stof: note that we should also add this dev dependency in our drivers, to make their own test fail when they use deprecated APIs (be it in Mink APIs or in Symfony ones btw). We don't want them to use deprecated APIs.

PR #663, opened 2015-08-29
Remove the dependency on the Session in expectation exceptions

stof:
Injecting the Session in exceptions is now deprecated in favor of injecting the driver.

This is a backport of https://github.com/minkphp/Mink/commit/761bb7965bf13dbb360ccace87e029fbb701489b which was done as part of #542 to refactor the driver in 2.0.
Backporting it with a BC layer means we give people instantiating our exception themselves (me at work for instance) an easier upgrade path, because they can prepare things before Mink 2.0.

### DIFF
--- a/src/Element/Element.php
+++ b/src/Element/Element.php
@@ -68,7 +68,7 @@ abstract class Element implements ElementInterface
      */
     public function getSession()
     {
-        @trigger_error(sprintf('The method %s is deprecated as of 1.7 and will be removed in 2.0', __METHOD__), E_USER_DEPRECATED);
+        @trigger_error(sprintf('The method %s is deprecated as of 1.6 and will be removed in 2.0', __METHOD__), E_USER_DEPRECATED);
 
         return $this->session;
     }
@@ -202,9 +202,13 @@ abstract class Element implements ElementInterface
      * @param string|null $locator
      *
      * @return ElementNotFoundException
+     *
+     * @deprecated as of 1.7, to be removed in 2.0
      */
     protected function elementNotFound($type, $selector = null, $locator = null)
     {
+        @trigger_error(sprintf('The method %s is deprecated as of 1.7 and will be removed in 2.0', __METHOD__), E_USER_DEPRECATED);
+
         return new ElementNotFoundException($this->driver, $type, $selector, $locator);
     }
 }
--- a/src/Element/Element.php
+++ b/src/Element/Element.php
@@ -197,9 +197,6 @@ abstract class Element implements ElementInterface
     /**
      * Builds an ElementNotFoundException
      *
-     * This is an helper to build the ElementNotFoundException without
-     * needing to use the deprecated getSession accessor in child classes.
-     *
      * @param string      $type
      * @param string|null $selector
      * @param string|null $locator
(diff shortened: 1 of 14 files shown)


src/Element/Element.php
```diff
@@ -204,6 +196,6 @@ public function getOuterHtml()
      */
     protected function elementNotFound($type, $selector = null, $locator = null)
```
stof: I'm wondering whether we should throw exceptions in place now and deprecate this method (this was added in 1.6 just so that our own child classes can avoid using the deprecated Session getter)

src/Exception/ExpectationException.php
```diff
@@ -22,17 +23,28 @@
...
+     * @param string                  $message   optional message
+     * @param DriverInterface|Session $driver    driver instance (or session for BC)
+     * @param \Exception|null         $exception expectation exception
      */
-    public function __construct($message, Session $session, \Exception $exception = null)
+    public function __construct($message, $driver, \Exception $exception = null)
     {
-        $this->session = $session;
+        if ($driver instanceof Session) {
+            @trigger_error('Passing a Session object to the ExpectationException constructor is deprecated as of Mink 1.7. Pass the driver instead.', E_USER_DEPRECATED);
```
aik099: What is the point of using error hiding (via `@`) if we want to trigger an error?

src/Exception/ExpectationException.php
```diff
@@ -75,6 +97,12 @@ protected function getContext()
      */
     protected function getSession()
     {
+        if (null === $this->session) {
+            throw new \LogicException(sprintf('The deprecated method %s cannot be used when passing a driver in the constructor', __METHOD__));
```
aik099: Exception message says, that method is deprecated, but we're missing `@deprecated` in the DocBlock.

src/WebAssert.php
```diff
@@ -417,7 +417,7 @@ public function elementExists($selectorType, $selector, ElementInterface $contai
                 $selector = implode(' ', $selector);
             }
 
-            throw new ElementNotFoundException($this->session, 'element', $selectorType, $selector);
+            throw new ElementNotFoundException($this->session->getDriver(), 'element', $selectorType, $selector);
```
aik099: Isn't `$this->driver` available in this class?

src/Element/Element.php
aik099: Isn't this just an internal method that contained some logic before, but not now? If so, then we'd better show exception where needed instead of calling method, that throws them for us. This would also give us shorter stacktrace for that exception.

stof: agreed. Precisely my point.
there was no logic before, but it allowed to access the session without using the deprecated getter

src/Exception/ExpectationException.php
stof: see #662

src/Exception/ExpectationException.php
stof: right. This is a mistake

src/WebAssert.php
stof: nope. this class uses the Session API.

src/Element/Element.php
stof: done

Issue #667, opened 2015-09-01
All repos should test against lowest deps, and maybe dev deps


stof:
We should have a job in the matrix testing against lowest dependencies to ensure that the lowest bound is right.

Depending on what they depend on, it might be worth testing against dev versions too (even though we always test them against the dev version of Mink).
You can see https://github.com/jcalderonzumba/MinkPhantomJSDriver/pull/36 for the way to make such setup (the PR contains other enhancements too)

aik099:
Interesting topic, but doing so we'll be testing the dependencies instead of library itself. If we believe SemVer, then testing against lowest version isn't needed.

stof:
no, we will be testing that our lowest bounds are right. Otherwise, it is very easy to say we are compatible with twig 1.0+ for instance, while we actually depend on a feature added in twig 1.8 (this is an example taken from the PR linked above).

And for the dev stability, this is meant to ensure we are compatible with the next version of our dependencies, to catch issues coming from their update (which should not happen if they follow semver, but this still happens sometimes by mistakes)

aik099:
Ah, so we test against lowest dependency version instead of latest dependency version. In that case yes, I agree, that it's needed.

stof:
@aik099 this is what ``--prefer-lowest`` is about in Composer

aik099:
I though you want to test lowest version together with highest versions :)

stof:
I don't understand what you mean ?

stof:
note that I sent the PR against the Mink repo already. PRs against drivers will come later

aik099:
You said: let's test against lowest version of dependencies.
I understood as: let's test against highest version of dependencies (as we do now) + test against lowest version of dependencies.
What you actually meant: let's test against lowest version of dependencies instead of highest version of dependencies.

stof:
no, we should do both. We should have 1 job using the lowest deps, but we should also have jobs using the latest ones (most of our users will actually use the latest ones, so it is important to use them).

aik099:
It's clear now. I was afraid that testing against lowest dependency version would double travis job count, but since it will only add one job, then it's ok.

stof:
yeah, testing lowest deps against all PHP versions would be overkill. We are not trying to test our deps on multiple versions (they should be doing it), and we are likely to execute the same code paths in our own code. What I generally do is using our older PHP version for the lowest deps build.

stof:
[...]

PR #668, opened 2015-09-01
Add testing against lowest dependencies

stof:
Refs #667

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,21 +2,27 @@ language: php
 
 sudo: false
 
-php: [5.3, 5.4, 5.5, 5.6, nightly, hhvm]
+php: [5.3, 5.4, 5.5, 5.6, 7.0, hhvm]
 
 matrix:
   fast_finish: true
+  include:
+    - php: 5.3
+      env: COMPOSER_FLAGS='--prefer-lowest --prefer-stable' SYMFONY_DEPRECATIONS_HELPER=weak
   allow_failures:
-    - php: nightly
+    - php: 7.0
 
 cache:
   directories:
     - $HOME/.composer/cache/files
 
+before_install:
+  - composer self-update
+
 install:
-  - composer install
+  - composer update $COMPOSER_FLAGS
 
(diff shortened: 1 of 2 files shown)


composer.json
```diff
@@ -16,7 +16,7 @@
 
     "require": {
         "php":                   ">=5.3.1",
-        "symfony/css-selector":  "~2.0"
+        "symfony/css-selector":  "~2.1"
```
stof: Early 2.0 components have broken autoloading configurating, making them unusable with Composer

.travis.yml
```diff
@@ -2,21 +2,27 @@ language: php
...
+
 install:
-  - composer install
+  - composer update $COMPOSER_FLAGS
 
 script: phpunit -v --coverage-clover=coverage.clover
 
 after_script:
-  - if [ "nightly" != "$TRAVIS_PHP_VERSION" ]; then wget https://scrutinizer-ci.com/ocular.phar && php ocular.phar code-coverage:upload --format=php-clover coverage.clover; fi
+  - if [[ "7.0" != "$TRAVIS_PHP_VERSION" && "$TRAVIS_PHP_VERSION" != "hhvm" ]]; then wget https://scrutinizer-ci.com/ocular.phar && php ocular.phar code-coverage:upload --format=php-clover coverage.clover; fi
```
stof: recent PHPUnit versions are not able to generate code coverage on the old HHVM 3.5 (which is still used by Travis), so I avoid notifying Scrutinizer of a missing coverage.

.travis.yml
```diff
@@ -2,21 +2,27 @@ language: php
 
 sudo: false
 
-php: [5.3, 5.4, 5.5, 5.6, nightly, hhvm]
+php: [5.3, 5.4, 5.5, 5.6, 7.0, hhvm]
```
stof: I switch to 7.0 instead of nightly, so that this can use 7.0 releases once they become available on Travis. currently, it is just an alias for nightly

PR #670, opened 2015-09-01
Add a test ensuring that drivers follow conventions

stof:
This ensures that drivers are implemented in a forward compatible way for Twig 2, and that they implement the most basic methods necessary to make the driver usable in Mink.

### DIFF
--- a/driver-testsuite/tests/Basic/BestPracticesTest.php
+++ b/driver-testsuite/tests/Basic/BestPracticesTest.php
@@ -1,0 +1,84 @@
+<?php
+
+namespace Behat\Mink\Tests\Driver\Basic;
+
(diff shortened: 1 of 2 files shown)


driver-testsuite/tests/Basic/BestPracticesTest.php
```diff
@@ -0,0 +1,84 @@
+<?php
+
+namespace Behat\Mink\Tests\Driver\Basic;
+
+use Behat\Mink\Tests\Driver\TestCase;
+
+/**
+ * This testcase ensures that the driver implementation follows recommended practices for drivers.
+ */
+class BestPracticesTest extends TestCase
```
stof: if you have a better idea for the class name, feed free to participate

aik099:
OK.

PR #671, opened 2015-09-02
The "NodeElement::hasClass" now checks for newlines as well

### LINKED ISSUE #657: The "NodeElement::hasClass" doesn't support new lines
In HTML it's allowed for css classes, listed in `class` html attribute to be separated by:

* any whitespace (not space specifically)
* any number of whitespace (not single one)

The `NodeElement::hasClass` method (see https://github.com/minkphp/Mink/blob/master/src/Element/NodeElement.php#L142) expects class names to be separated by single space. Therefore in the following CSS classes won't be properly detected:

```html
<div class="
one
two

three
"></div>
```
Yeah, some frameworks do generate such HTML.

aik099:
Closes #657

### DIFF
--- a/tests/Element/NodeElementTest.php
+++ b/tests/Element/NodeElementTest.php
@@ -156,11 +156,13 @@ class NodeElementTest extends ElementTest
             ->expects($this->exactly(6))
             ->method('getAttribute')
             ->with('input_tag', 'class')
-            ->will($this->returnValue('class1 class2'));
+            ->will($this->returnValue('
+            class1  class2
+            '));
 
-        $this->assertTrue($node->hasClass('class1'));
-        $this->assertTrue($node->hasClass('class2'));
-        $this->assertFalse($node->hasClass('class3'));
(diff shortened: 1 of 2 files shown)


tests/Element/NodeElementTest.php
```diff
@@ -156,11 +156,13 @@ public function testHasClass()
...
+            ->will($this->returnValue('
+            class1  class2
+            '));
 
-        $this->assertTrue($node->hasClass('class1'));
-        $this->assertTrue($node->hasClass('class2'));
-        $this->assertFalse($node->hasClass('class3'));
+        $this->assertTrue($node->hasClass('class1'), 'The "class1" was found');
+        $this->assertTrue($node->hasClass('class2'), 'The "class2" was found');
+        $this->assertFalse($node->hasClass('class3'), 'The "class3" was found');
```
stof: the message should not be the same than above as the expectation is not the same

aik099: Fixed.

PR #676, opened 2015-10-05
Add testing against dev versions of dependencies

stof:
This adds a job in the matrix running tests against dev versions of our dependencies.
The goal is to allow us to catch failures on them before the new version of the dependency gets released (assuming we have a Travis build running of course), allowing us to prepare changes before the stable release of our dependency. This is mostly useful to catch places where the APIs we are using are getting deprecated.


stof:
The second commit is fixing the issue with the CssSelector to use the non-deprecated API in Symfony 2.8+

aik099:
OK.

dawehner:
Looks RTBM for me. Thank you @stof

PR #675, opened 2015-10-04
Add a test for cookies with semicolon

dawehner:
In integration between Drupal and the phantomjs driver in mink (https://github.com/jcalderonzumba/MinkPhantomJSDriver)
we ran into an issue with cookies containing semicolons. This bug is probably due to missing url encoding in phantomjs itself
or some JS.

This PR is about ensuring that using semicolons actually work.

PR #680, opened 2015-10-22
Revert "Add a test for cookies with semicolon"

aik099:
Reverts minkphp/Mink#675 because the semicolon isn't allowed symbol in cookie value. Quote from http://curl.haxx.se/rfc/cookie_spec.html : 

```
NAME=VALUE
This string is a sequence of characters excluding semi-colon, comma and white space. If there is a need to place such data in the name or value, some encoding method such as URL style %XX encoding is recommended, though no encoding is defined or required.
```

// cc: @dawehner

### DIFF
--- a/driver-testsuite/tests/Basic/CookieTest.php
+++ b/driver-testsuite/tests/Basic/CookieTest.php
@@ -47,14 +47,6 @@ class CookieTest extends TestCase
         $this->assertContains('Previous cookie: NO', $this->getSession()->getPage()->getText());
     }
 
-    public function testCookieWithSemicolon()
-    {
-        $this->getSession()->setCookie('srvr_cookie', 'foo;bar;baz');
-        $this->getSession()->visit($this->pathTo('/cookie_page2.php'));
-        $this->assertEquals('foo;bar;baz', $this->getSession()->getCookie('srvr_cookie'));
-        $this->assertContains('Previous cookie: foo;bar;baz', $this->getSession()->getPage()->getText());
-    }
-
     /**
      * @dataProvider cookieWithPathsDataProvider
      */


aik099:
I've used `Revert` button on PR page. Who knew that it would create PR branch in origin repo and not my fork. Sorry.

stof:
I'm wondering whether we should consider that Mink expects a raw cookie value, or a cookie value it will encode. the ``setcookie`` function in PHP is encoding the value first, so most people probably assume that the cookie API deals with encoding for them.

aik099:
So the following PHP code is possible:
```php
setcookie('test', 'a;b');
```
Then what `Set-Cookie` header will that produce?

I thought that cookie is present in header as-is.

aik099:
If code above works (the `;` in `setcookie` call), then we should update each driver to now auto-encode (I wonder what encoding function should be used) given cookie value from user before sending it elsewhere (e.g. WebDriver, Zombie, etc.). That would be considered as BC break however.

stof:
@aik099 see http://php.net/setcookie vs http://php.net/setrawcookie

stof:
And ``$_COOKIE`` contains url-decoded cookies, so PHP developers are seeing raw cookies only if they parse headers manually

aik099:
So you're saying that we can do `urlencode($user_supplied_cookie);` in drivers (if we're not afraid of BC break) and all should be fine.

Then this reverting PR should closed I guess.

dawehner:
Thank you @aik099 and @stof for discussing this out.
So yes indeed, with the original PR I wanted to make it behave as PHP, which was the reason why the code which was broken by that, used ";" originally

aik099:
@dawehner , could you please create PR for each driver to escape given cookie name/value?

dawehner:
Totally forgot about that, sorry

stof:
I triggered a build for all maintained driver a few days ago, so you can see their latest Travis build to see whether they pass this test or no

aik099:
@dawehner are PR's coming?

dawehner:
I hope so, but I can't promise anything. This notification at least is
marked as "unread" in my mail client

stof:
Closing in favor of https://github.com/minkphp/Mink/pull/686 which fixes the test to use the Mink API properly. I will then fix drivers needing a fix.

aik099:
OK.

PR #687, opened 2016-01-26
Fix the compatibility of the testsuite with PHPUnit 5

### LINKED ISSUE #684: Fatal Error, when running driver test suite on PHP 7
At least MinkZombieDriver build fails due Fatal Error on PHP 7 (see https://travis-ci.org/minkphp/MinkZombieDriver/jobs/102687778):

```
PHP Fatal error:  Uncaught Declaration of Behat\Mink\Tests\Driver\TestCase::onNotSuccessfulTest(Exception $e) should be compatible with PHPUnit_Framework_TestCase::onNotSuccessfulTest($e)
```

The signature on PhpUnit used for PHP 5:

```php
protected function onNotSuccessfulTest(Exception $e)
```

The signature on PhpUnit used for PHP 7:

```php
protected function onNotSuccessfulTest($e)
```

Our overridden method signature:

```php
protected function onNotSuccessfulTest(\Exception $e)
```

stof:
I created a separate class holding this logic, to avoid duplicating the whole base TestCase class, thus reducing duplication.

Closes #684

### DIFF
--- a/driver-testsuite/tests/SkippingUnsupportedTestCase.php
+++ b/driver-testsuite/tests/SkippingUnsupportedTestCase.php
@@ -1,0 +1,44 @@
+<?php
+
+namespace Behat\Mink\Tests\Driver;
+
+use Behat\Mink\Exception\UnsupportedDriverActionException;
+
+if (version_compare(\PHPUnit_Runner_Version::id(), '5.0.0', '>=')) {
+    /**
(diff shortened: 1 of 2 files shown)


stof:
I will ignore the Scrutinizer issues being reported here (I know I define the same class twice, as this is even the goal of this PR). I'm merging this to be able to work on drivers properly.

PR #691, opened 2016-02-24
Export-ignore test files

barryvdh:
This will exclude the tests en driver-testsuite (+ some other files) from the archives, downloaded with Composer and directly via Github. As this code is not used in production, it doesn't have to be included.

People can still use --prefer-source to get the full archive with tests etc. Should also prevent code from being run when the vendor folder is not properly protected (#677)

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,0 +1,8 @@
+/driver-testsuite export-ignore
+/tests export-ignore
+.gitattributes export-ignore
+.gitignore export-ignore
+.travis.yml export-ignore
+CONTRIBUTING.md export-ignore
+phpdoc.ini.dist export-ignore
+phpunit.xml.dist export-ignore
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,4 +1,3 @@
-/driver-testsuite export-ignore
 /tests export-ignore
 .gitattributes export-ignore
 .gitignore export-ignore


aik099:
The driver repos are using Mink as dependency and are running `driver-testsuite`. Won't this change prevent them from working? I suppose we need to make sure `--prefer-source` option is used in all driver `.travis.yml` files.

barryvdh:
Sorry, I don't really know the internals of Mink, so perhaps there are better solutions. I would just like to suggest to package only what is needed.

If other packages are relying on the driver-testsuite for their tests, wouldn't it be better to create a seperate repository for the driver-testsuite? You can require-dev that package from here and the other drivers.

barryvdh:
But yeah, I don't know how many drivers there are, and if they're all maintained by you, but adding --prefer-source to the Travis script would be a simple solution also :)

aik099:
These are the all drivers maintained by Mink: https://github.com/minkphp?utf8=%E2%9C%93&query=Driver
This are 3rd party drivers that I'm aware of: https://github.com/jcalderonzumba/MinkPhantomJSDriver

.gitattributes
```diff
@@ -0,0 +1,8 @@
+/driver-testsuite export-ignore
```
stof: This folder is necessary when installing Mink as a dependency of a driver, as it is a testsuite meant to be used by drivers

barryvdh: See comment here: https://github.com/minkphp/Mink/pull/691#issuecomment-188154543

> If other packages are relying on the driver-testsuite for their tests, wouldn't it be better to create a seperate repository for the driver-testsuite? You can require-dev that package from here and the other drivers.

stof: Well, the issue with this setup would be versionning for changes in Mink. But yeah, this is something I already thought about.

barryvdh: Yes and the alternative is using --prefer-source in the other drivers travis setup.

stof: not only on Travis but for anyone working on drivers, making contributing harder. And forcing to use source installs is also bad as it means that all other deps of the driver will also be installed from source, forbidding to use the cache (making builds slower)

barryvdh: Well, the builds mostly aren't cached anyways because of API limits (they're cloning as fallback): https://travis-ci.org/minkphp/MinkGoutteDriver/jobs/98220899

But agreed, it would make it harder for them. But it seems a bit odd to include a development dependancy for everyone. As a seperate repository, it wouldn't have to be included for the rest of the users (the 99%)

stof: Well, the cache contains older archives. At one point, we were lucky to avoid hitting the rate limit.

I totally agree that a separate repo would be better for this particular reason. But it is not done yet, so excluding it from the archive today is wrong.

barryvdh: I've removed that line.

.gitattributes
```diff
@@ -0,0 +1,7 @@
+/tests export-ignore
```
aik099: Isn't this folder somehow used, when running driver test suite as well?

barryvdh: Is that possible, when it's only autoloaded on dev?

aik099: I've double checked with one of the drivers (see https://github.com/minkphp/MinkSelenium2Driver/blob/master/phpunit.xml.dist#L7) and it seems, that only `driver-testsuite` folder is included in PHPUnit. So this should work theoretically.

stof: I confirm that the driver testsuite does not need this at all.

stof:
OK, merging this to make the archive smaller. The case of the driver-testsuite will be handled separately

PR #694, opened 2016-02-27
Fixes XSS in the driver test suite

aik099:
I've tested on all drivers except Sahi, which test suite takes too much time to run in any case.

Another PR for MinkBrowserKit driver would be sent, because it overrides fixture files that were fixed in main test suite.

Replaces #677

### DIFF
--- a/driver-testsuite/web-fixtures/utils.php
+++ b/driver-testsuite/web-fixtures/utils.php
@@ -1,0 +1,16 @@
+<?php
+
+function html_escape_value($data)
+{
+    if (!is_array($data)) {
+        return htmlspecialchars($data, ENT_QUOTES, 'UTF-8', false);
+    }
+
+    $escapedData = array();
+
+    foreach ($data as $key => $value) {
+        $escapedData[html_escape_value($key)] = html_escape_value($value);
+    }
+
+    return $escapedData;
+}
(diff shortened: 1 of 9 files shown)


driver-testsuite/web-fixtures/utils.php
```diff
@@ -0,0 +1,16 @@
+<?php
+
+function htmlspecialchars_array($data)
+{
+    if (!is_array($data)) {
+        return htmlspecialchars($data, ENT_QUOTES, 'UTF-8', false);
+    }
+
+    $escaped_data = array();
```
stof: this should use a camelCased name

driver-testsuite/web-fixtures/utils.php
```diff
@@ -0,0 +1,16 @@
+<?php
+
+function htmlspecialchars_array($data)
```
stof: the name is misleading IMO, as it handles more than arrays

aik099: Anything specific you have in mind?

driver-testsuite/web-fixtures/utils.php
aik099: Fixed.

driver-testsuite/web-fixtures/utils.php
stof: ``html_escape_value`` ?

stof:
@aik099 I was about to release a new version of Mink, but I think this may be merged before the release.

driver-testsuite/web-fixtures/utils.php
aik099: Will do.

aik099: Change pushed.

PR #695, opened 2016-03-04
Update the changelog for 1.7.1

stof:
This is preparing the release.

I'm considering including #694 in the release too though, if @aik099 finishes it.


stof:
Note that I want to tag Mink before tagging drivers (BrowserKitDriver will need to increase the min version after the merge of #694 and its BrowserKitDriver counterpart)

aik099:
I've renamed the `htmlspecialchars_array` into `html_escape_value` in #694 as you've requested.

aik099:
I think, that we should specially mention that XSS fix was made, because current message in change log `* Added HTML escaping of submitted values in the driver testsuite web-fixtures` doesn't really tell that.

Usually security-related items in changelog should be separated in it's own group.

PR #697, opened 2016-03-12
Fix WebAssert::fieldValue(Not)Equals

lauft:
Behat/Mink/WebAssert.php, `fieldValueEquals`/`fieldValueNotEquals`:
- Replace usage of `preg_match` by `===`
- Extend comparison ability to arrays
- Add array handling to message creation
- Comparison for boolean values seems broken, but booleans are only returned by checkboxes which are more likely to be checked by `WebAssert::checkboxChecked`/`WebAssert::checkboxNotChecked`

tests/WebAssertTest.php:
- Add special handling for `PHPUnit_Framework_AssertionFailedError`:
-- PHPUnit_Framework_Assert::fail throws this exception and the test should let it pass
-- Test should catch and process all other exception
- Fix and extend tests `testFieldValueEquals` and `testFieldValueNotEquals`:
-- according to `PHPDoc Behat\Mink\Element\NodeElement::getValue` returns `string|bool|array`
-- current tests only check against integer values

### DIFF
--- a/src/Behat/Mink/WebAssert.php
+++ b/src/Behat/Mink/WebAssert.php
@@ -558,7 +558,7 @@ class WebAssert
     }
 
     /**
-     * Checks that specific field have provided value.
+     * Checks that specific field value equals provided value.
      *
      * @param string             $field     field id|name|label|value
      * @param string             $value     field value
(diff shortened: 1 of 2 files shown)


aik099:
Please rebase. Since you've created your PR initially based on past version of Mink it's conflicting with current Mink version now.

PR #699, opened 2016-03-21
Fix typo

funivan:
Fix trigger error methods call. 
1. Someone forgot to pass `__METHOD__` to `sprintf` function
2. Error type should be `E_USER_DEPRECATED` instead of notice


stof:
thanks

PR #701, opened 2016-04-15
Use selectOption if setValue is called on select element

vaidasm:
This allows to use option labels when using the following syntax (currently this works with option value only):
```
  When I fill in the following:
    | Example label      | Example label  |
```

### DIFF
--- a/src/Element/TraversableElement.php
+++ b/src/Element/TraversableElement.php
@@ -157,7 +157,11 @@ abstract class TraversableElement extends Element
             throw new ElementNotFoundException($this->getDriver(), 'form field', 'id|name|label|value|placeholder', $locator);
         }
 
-        $field->setValue($value);
+        if ('select' === $field->getTagName()) {
+            $field->selectOption($value);
+        } else {
+            $field->setValue($value);
+        }
     }
 
     /**
--- a/tests/Element/DocumentElementTest.php
+++ b/tests/Element/DocumentElementTest.php
@@ -318,6 +318,31 @@ class DocumentElementTest extends ElementTest
         $this->document->fillField('some field', 'some val');
     }
 
+    public function testSetValueOnSelect()
+    {
+        $node = $this->getMockBuilder('Behat\Mink\Element\NodeElement')
+            ->disableOriginalConstructor()
+            ->getMock();
+        $node
+            ->expects($this->once())
+            ->method('getTagName')
+            ->willReturn('select');
+        $node
+            ->expects($this->once())
+            ->method('selectOption')
+            ->with('some val');
+
+        $this->mockNamedFinder(
+            '//field',
+            array($node),
+            array('field', 'some field')
+        );
+
+        $this->document->fillField('some field', 'some val');
+        $this->setExpectedException('Behat\Mink\Exception\ElementNotFoundException');
+        $this->document->fillField('some field', 'some val');
+    }
+
     public function testCheckField()
     {
         $node = $this->getMockBuilder('Behat\Mink\Element\NodeElement')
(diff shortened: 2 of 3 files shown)


stof:
Tests must be added to cover this behavior:

- a unit test in the Mink testsuite
- a functional test in the driver testsuite

src/Element/TraversableElement.php
```diff
@@ -157,7 +157,11 @@ public function fillField($locator, $value)
             throw new ElementNotFoundException($this->getDriver(), 'form field', 'id|name|label|value|placeholder', $locator);
         }
 
-        $field->setValue($value);
+        if ($field->getTagName() === "select") {
```
aik099: Please replace double quotes with single quotes.

aik099: Doesn't `setValue` in the drivers cover this case already?

aik099: I've found https://github.com/minkphp/MinkSelenium2Driver/blob/master/src/Selenium2Driver.php#L612-L626 code for Selenium2Driver where this is already done. Not sure if other drivers do that as well.

stof: well, they should, which is precisely why I said that the testsuite needs to cover it, to see whether the functional test passes without the change when running it in drivers

vaidasm: BrowserKit implementation is more straightforward:

https://github.com/minkphp/MinkBrowserKitDriver/blob/master/src/BrowserKitDriver.php#L414-L417

So is this browserkit driver bug? Should this be fixed in there or..?

aik099: As @stof suggested you need to create test in this PR that we'll run on all drivers and see if additional PRs to fix the drivers are needed.

aik099: > BrowserKit implementation is more straightforward: https://github.com/minkphp/MinkBrowserKitDriver/blob/master/src/BrowserKitDriver.php#L414-L417

What you've specified is driver implementation. This is because the BrowserKit's `setValue` method is clever enough to understand how value should be set into the fields.

aik099: Actually it's DomCrawler who is doing all the setting: https://github.com/symfony/dom-crawler/tree/master/Field

vaidasm: @aik099 I updated the pull request with the unit and driver test. Also
https://github.com/symfony/dom-crawler/blob/master/Field/ChoiceFormField.php#L120

does not accept label as an argument, only values. So which is it? driver should accept label as valid argument for `setValue` or Mink should use selectOption on 'select' nodes?

stof:
Actually, I'm not convinced that ``setValue`` should accept the option label as value. This will *not* be the value being set.

I think you should just use ``selectOption`` in your code when you want this behavior.

src/Element/TraversableElement.php
```diff
@@ -157,7 +157,11 @@ public function fillField($locator, $value)
             throw new ElementNotFoundException($this->getDriver(), 'form field', 'id|name|label|value|placeholder', $locator);
         }
 
-        $field->setValue($value);
+        if ('select' === $field->getTagName()) {
```
aik099: This change should be reverted, because it's driver responsibility.

tests/Element/DocumentElementTest.php
```diff
@@ -318,6 +318,31 @@ public function testFillField()
         $this->document->fillField('some field', 'some val');
     }
 
+    public function testSetValueOnSelect()
```
aik099: This should be driver-level test, not NodeElement-level test.

PR #703, opened 2016-05-06
Fix piped selectors for named elemends

### LINKED ISSUE #702: Cannot fill in fields with pipes (|) in name-attribute
I want to fill in field with pipes in the name a value, like

<code>And I fill in "a|b|c" with "1"</code>

The HTML form-element looks like:
```html
<input type="text" name="a|b|c" value="" />
```

The [Manipulator] splits the name-attribute.  May be the [NamedSelector] should escape the pipe-signs.

[Manipulator]: https://github.com/minkphp/Mink/blob/master/src/Selector/Xpath/Manipulator.php#L24
[NamedSelector]: https://github.com/minkphp/Mink/blob/master/src/Selector/NamedSelector.php#L244-L262

gabel:
Fixes #702 

Fixes non working sentence:

```gherkin
Feature: fill in fields with pipes (|) in name-attribute

  Scenario:
    Given I am on "/test.html"
    When I fill in "a|b|c" with "foobar" #  the name selector could not be found
    Then the "a|b|c" field should contain "foobar"
```

Non working context
````html
<!DOCTYPE html>
<html lang="en">
<head>
    <meta charset="UTF-8">
    <title>Title</title>
</head>
<body>
<form>
    <input type="text" name="a|b|c" />
</form>

</body>
</html>
```

f the fix is OK, I'll add tests for this specific behavior.

### DIFF
--- a/src/Selector/Xpath/Manipulator.php
+++ b/src/Selector/Xpath/Manipulator.php
@@ -21,7 +21,7 @@ class Manipulator
     /**
      * Regex to find union operators not inside brackets.
      */
-    const UNION_PATTERN = '/\|(?![^\[]*\])/';
+    const UNION_PATTERN = '/\s\|\s(?![^\[]*\])/';
 
     /**
      * Prepends the XPath prefix to the given XPath.


stof:
As commented in https://github.com/gabel/Mink/pull/1#issuecomment-217439172, this is totally broken (and adding tests reproducing the bug before trying to implement the fix would have shown this immediately btw)

gabel:
In my op this is a matter of interpretation. Surely xpath does allow `a|b` but without major changes the Regex in [Manipulator](https://github.com/minkphp/Mink/blob/master/src/Selector/Xpath/Manipulator.php#L24) wont solve the issue #702.

stof:
I know that the Manipulator has a bug today. but your change does not fix it. It replaces it with 2 bugs:

- one being the same than the current bug, but affecting a few less cases
- one breaking XPath queries which are working today

So this PR won't be accepted. We need to have a fix (which needs to make the Manipulator smarter about what XPath is), not a hack breaking other things.

gabel:
You're right but I'm not able to fix this issue for now.. :cry:

stof:
Closing in favor of #720

PR #705, opened 2016-06-17
Auto-start session only upon 1st "->visit(...)" method call

### LINKED ISSUE #678: Auto-start session on "->visit(...)" call
Currently if somebody attempts to do a `$session->visit(...);` call without calling `$session->start()` upfront, then this will ends up in exception.

To help out users I propose to auto-start session (only if not started already) inside `$session->visit(...)` method before any existing code.

aik099:
Closes #678

### DIFF
--- a/src/Mink.php
+++ b/src/Mink.php
@@ -100,7 +100,7 @@ class Mink
     }
 
     /**
-     * Returns registered session by it's name or active one and automatically starts it if required.
+     * Returns registered session by it's name or default one.
      *
      * @param string $name session name
      *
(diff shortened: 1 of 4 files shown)


aik099:
@stof , any updates on this?

stof:
There is a difference now: ``assertSession()`` may be called for a non-started session, while it was starting it previously. Should we force starting ? It would match the previous behavior, but I'm not sure it is necessary. The assertions are not usable without visiting a page anyway.

aik099:
None of `WebAssert` class methods are calling `visit` internally. I suppose the `WebAssert` class is used something like this:

```php
$session = ...;// obtain session somehow
$session->visit('page_url');
$assert = $mink->assertSession($session);
$assert->assert...
```
If that is the case, then no changes should be made to `Mink::assertSession` because the started session won't be usable anyway without visiting a page even though technically it's a BC break.

stof:
yeah, even though it is technically a BC break, it is only breaking BC for unsupported usages of the API. So fine.

PR #708, opened 2016-07-01
Added doesNotHaveAttribute() and elementAttributeNotExists() with tests

J7mbo:
Should be really simple - it's just a few changes with negations. If you don't like the exception stuff however we can refactor that.

### DIFF
--- a/src/Element/NodeElement.php
+++ b/src/Element/NodeElement.php
@@ -118,18 +118,6 @@ class NodeElement extends TraversableElement
     }
 
     /**
-     * Checks whether element does not have an attribute with specified name.
-     *
-     * @param string $name
-     *
-     * @return Boolean
-     */
-    public function doesNotHaveAttribute($name)
-    {
-        return null === $this->getDriver()->getAttribute($this->getXpath(), $name);
-    }
-
-    /**
      * Returns specified attribute value.
      *
      * @param string $name
--- a/src/Element/NodeElement.php
+++ b/src/Element/NodeElement.php
@@ -118,6 +118,18 @@ class NodeElement extends TraversableElement
     }
 
     /**
+     * Checks whether element does not have an attribute with specified name.
+     *
+     * @param string $name
+     *
+     * @return Boolean
+     */
+    public function doesNotHaveAttribute($name)
+    {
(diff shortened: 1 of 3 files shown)


src/Element/NodeElement.php
```diff
@@ -118,6 +118,18 @@ public function hasAttribute($name)
     }
 
     /**
+     * Checks whether element does not have an attribute with specified name.
+     *
+     * @param string $name
+     *
+     * @return Boolean
+     */
+    public function doesNotHaveAttribute($name)
```
stof: -1 for this method. Use ``!$element->hasAttribute($name)`` in the WebAssert class

tests/WebAssertTest.php
```diff
@@ -982,6 +982,81 @@ public function testElementAttributeExists()
         );
     }
 
+    public function testElementAttributeNotExistsValid()
+    {
+        $page = $this->getMockBuilder('Behat\\Mink\\Element\\DocumentElement')
+                     ->disableOriginalConstructor()
+                     ->getMock()
```
stof: wrong indentation

tests/WebAssertTest.php
```diff
@@ -982,6 +982,81 @@ public function testElementAttributeExists()
...
+        ;
+
+        $element
+            ->expects($this->exactly(1))
+            ->method('doesNotHaveAttribute')
+            ->with('blah')
+            ->will($this->returnValue(true))
+        ;
+
+        call_user_func_array(array($this->assert, 'elementAttributeNotExists'), array('css', 'h2 > span', 'blah'));
```
stof: no need to use ``call_user_func_array`` here

stof: btw, it would be better to sue the existing ``assertCorrectAssertion`` and ``assertWrongAssertion`` methods instead

J7mbo: I was thinking of adding an extra parameter for the above methods to take into account this. What do you think, any ideas on that?

J7mbo:
I really can't fix the indentation here :-/

aik099:
You again have sent PR from master branch. You should close this one and send one from different branch.

PR #710, opened 2016-07-10
Tests for "DriverInterface::setValue" on SELECT by option name

spolischook:


spolischook:
https://github.com/minkphp/Mink/pull/710

driver-testsuite/tests/Form/SelectTest.php
```diff
@@ -91,6 +91,15 @@ public function testSetValueSingleSelect()
         $this->assertEquals('10', $select->getValue());
     }
 
+    public function testSetValueByDriverSingleSelect()
```
aik099: Please rename method into `testSetValueByOptionText`.

driver-testsuite/tests/Form/SelectTest.php
```diff
@@ -91,6 +91,15 @@ public function testSetValueSingleSelect()
...
+    {
+        $session = $this->getSession();
+        $session->visit($this->pathTo('/multiselect_form.html'));
+        $select = $session->getPage()->find('xpath', '//select[@name="select_number"]');
+        $select->setValue('thirty');
```
aik099: You should also add test (if such test is missing), that `select by option name` behavior works for multi-selects.

driver-testsuite/tests/Form/SelectTest.php
spolischook: done

driver-testsuite/tests/Form/SelectTest.php
```diff
@@ -102,6 +102,28 @@ public function testSetValueMultiSelect()
     }
 
     /**
+     * Set values by Driver
+     */
+    public function testSetValueByOptionText()
```
aik099: Please rewrite this test to use data providers. This would work like this:

1. data provider defines:
 * select location xpath
 * set values (for `setValue` call)
 * expected got values (from `getValue` call)
2. actual test would accept data provider data from method parameters and do a single assertion.

The change is needed for cases, when 1st assertion fails and it won't be visible that 2nd and 3rd might be passing.

spolischook: Good point, thanx!
Done.

driver-testsuite/tests/Form/SelectTest.php
```diff
@@ -102,6 +102,43 @@ public function testSetValueMultiSelect()
...
+
+        $select->setValue($value);
+        $this->assertEquals($expectedResult, $select->getValue());
+    }
+
+    public function setValueByOptionTextProvider()
+    {
+        return array(
+            'single select' => array(
+                'xpath'          => '//select[@name="select_number"]',
```
aik099: 1. Indentation in arrays isn't needed.
2. Key names aren't used by PHPUnit and can be removed.

driver-testsuite/tests/Form/SelectTest.php
```diff
@@ -102,6 +102,43 @@ public function testSetValueMultiSelect()
...
+        $session = $this->getSession();
+        $session->visit($this->pathTo('/multiselect_form.html'));
+        $page = $session->getPage();
+        $select = $page->find('xpath', $xpath);
+
+        $select->setValue($value);
+        $this->assertEquals($expectedResult, $select->getValue());
+    }
+
+    public function setValueByOptionTextProvider()
```
aik099: Please rename into `setValueByOptionTextDataProvider` (added `Data` word). Also change name where it's used in test itself.

driver-testsuite/tests/Form/SelectTest.php
spolischook: 1. ok
2. in most cases it easy to understand what data will be return by provider. But it's ok, if it not used in Mink, I removed it.

spolischook: done

aik099: done & pushed? I see no changes.

spolischook: yep, see https://github.com/minkphp/Mink/pull/710/files

driver-testsuite/tests/Form/SelectTest.php
spolischook: done

driver-testsuite/tests/Form/SelectTest.php
spolischook: changed by this commit https://github.com/minkphp/Mink/pull/710/commits/95688dbad96d2d65ca68da15844ffc610aa8eddb

stof:
Have you tried running this test against the different drivers ? If yes, what is the result ?

spolischook:
@stof Run on MinkSelenium2Driver. 
On master branch it was fail, but with fix https://github.com/minkphp/MinkSelenium2Driver/pull/240 it will green.
Should I test any other drivers?

aik099:
> Should I test any other drivers?

Yes, all of them.

stof:
Would be great to test BrowserKitDriver (easy as it is pure PHP and so you just need to perform a ``composer install`` to be ready to run the testsuite, and then apply this patch to the testsuite), and ZombieDriver.
No need to test SeleniumDriver as it is deprecated and SahiDriver (as we also plan to deprecate it).
GoutteDriver uses the code of BrowserKitDriver. For such change, it will work as soon as BrowserKitDriver supports it.

stof:
The question then is whether we want to support it. Using ``setValue()`` to set something else than the value looks a bit weird to me.

stof:
Maybe selecting an option based on its text should be a different method, as this could be ambiguous: what if an option is matched by value and the other one by text ? Your change would break BC there (selecting a different option depending on their order)

spolischook:
MinkBrowserKitDriver
```
1) Behat\Mink\Tests\Driver\Form\SelectTest::testSetValueByOptionText with data set #0 ('//select[@name="select_number"]', 'thirty', '30')
InvalidArgumentException: Input "select_number" cannot take "thirty" as a value (possible values: 10, 20, 30).

vendor/symfony/dom-crawler/Field/ChoiceFormField.php:140
src/BrowserKitDriver.php:416
vendor/behat/mink/src/Element/NodeElement.php:105
vendor/behat/mink/driver-testsuite/tests/Form/SelectTest.php:116
```

It really strange behavior for me, because if I wrote BDD test I guided by option text, instead of value that I will not see on the form element.

aik099:
As I said before: some drivers specifically want to have option value, not title in the `setValue` call. The underlying component used by BrowserKit driver is one of them I guess. Probably special handling (if possible at all) needs to be done to select option by it's title.

spolischook:
@stof @aik099 
see this changes for drivers:
- https://github.com/minkphp/MinkSelenium2Driver/pull/240
- https://github.com/minkphp/MinkBrowserKitDriver/pull/110
- https://github.com/minkphp/MinkZombieDriver/pull/173
I was test them with this tests. All cases from dataProvider are green.

spolischook:
> what if an option is matched by value and the other one by text ?

Can't imagine this test case, can you help me with this?

> Maybe selecting an option based on its text should be a different method, as this could be ambiguous

this would be more ambiguous if element will have two methods for same action.
How ever I would be happy to see any suggestions about this changes.

aik099:
> Can't imagine this test case, can you help me with this?

Easy:

```html
<select name="test">
  <option value="one">one option</option>
  <option value="two">one</option>
```

> this would be more ambiguous if element will have two methods for same action.

I see, that `NodeElement::selectOption` already works with both option value and title. Then why we need the `DriverInterface::setValue` to support it?

stof:
> this would be more ambiguous if element will have two methods for same action.

My point is that it is **not** the same action. When using the option text rather than the value, you are **not** setting the value.

spolischook:
Ok. First of all thank you guys for your attention, you do a good work and build nice library.

My problem in my own code. I have code for fill all my forms. Simplified, it looks like:

```php
use Behat\Mink\Element\NodeElement;

class Form extends Element
{
    /**
     * @param TableNode $table
     * @throws ElementNotFoundException
     */
    public function fill(TableNode $table)
    {
        foreach ($table->getRows() as $row) {
            $locator = $row[0];
            $value   = $row[1];
            $this->fillField($locator, $value);
        }
    }
}
```

Now I can determine what field is, and set it properly. But the same work do Selenium2Driver [in setValue method](https://git.io/vK8LQ).
So my question is - who is responsible to determine field type and set it properly?

Another problem is that I'm missing ability to extend Element node that [was created by CoreDriver in find method](https://git.io/vK8Lu). If I can return my own type of element it would be easier to extend element behavior. 
So it would be great when I ask the driver (or element, any way it request the Driver) to find element it will returns the specific element (e.g. SelectFormElement, TextInputFormElement, TinyMceFormElement) that have additional methods and know how to fill itself.
What you think about it?

stof:
@spolischook in Mink 1.6, we added support for any field type in ``setValue``. But this still expect to set the value. In your case, you are expecting to fill the field with something else than the value, so it is an issue in your code IMO.

Regarding your request to extend the Element object, this is a bad idea IMO. Inheritance is the worst extension point to maintain regarding BC (any method we would add could break your child class which adds its own methods). I suggest you to use composition rather than inheritance instead.

aik099:
> Another problem is that I'm missing ability to extend Element node that was created by CoreDriver in find method. If I can return my own type of element it would be easier to extend element behavior. 
> So it would be great when I ask the driver (or element, any way it request the Driver) to find element it will returns the specific element (e.g. SelectFormElement, TextInputFormElement, TinyMceFormElement) that have additional methods and know how to fill itself.

For that purpose there is https://github.com/qa-tools/qa-tools project, that uses Mink and indeed creates more specialized versions of NodeElement.

aik099:
For example `Select` class for working with dropdown (see https://github.com/qa-tools/qa-tools/blob/master/library/QATools/QATools/HtmlElements/Element/Select.php) has dedicated methods for selecting options by value and by text specifically. No guessing.

stof:
I'm closing this, because I don't think allowing ``setValue`` to set something else than the value is a good idea, as explained previously.
And the PR is not mergeable anyway, as the driver testsuite has moved.

PR #711, opened 2016-07-17
Provide partial named selector that selects also the hidden fields

claudiu-cristea:
Sometimes we have to search for hidden fields (`<input type="hidden" .../>`). Add this selector to be used for such cases.

### DIFF
--- a/src/Selector/PartialNamedIncludingHiddenSelector.php
+++ b/src/Selector/PartialNamedIncludingHiddenSelector.php
@@ -1,0 +1,26 @@
+<?php
+
+/*
+ * This file is part of the Mink package.
+ * (c) Konstantin Kudryashov <ever.zet@gmail.com>
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+namespace Behat\Mink\Selector;
+
+/**
+ * Named selectors engine. Uses registered XPath selectors to create new expressions.
+ *
+ * @author Konstantin Kudryashov <ever.zet@gmail.com>
+ */
+class PartialNamedIncludingHiddenSelector extends PartialNamedSelector
+{
+    public function __construct()
+    {
+        $this->registerReplacement('%notFieldTypeFilter%', "not(%buttonTypeFilter%)");
+
+        parent::__construct();
+    }
+}


src/Selector/PartialNamedIncludingHiddenSelector.php
```diff
@@ -0,0 +1,26 @@
...
+/**
+ * Named selectors engine. Uses registered XPath selectors to create new expressions.
+ *
+ * @author Konstantin Kudryashov <ever.zet@gmail.com>
+ */
+class PartialNamedIncludingHiddenSelector extends PartialNamedSelector
+{
+    public function __construct()
+    {
+        $this->registerReplacement('%notFieldTypeFilter%', "not(%buttonTypeFilter%)");
```
aik099: Name doesn't really say that hidden fields are allowed.

src/Selector/PartialNamedIncludingHiddenSelector.php
```diff
@@ -0,0 +1,26 @@
...
+ */
+
+namespace Behat\Mink\Selector;
+
+/**
+ * Named selectors engine. Uses registered XPath selectors to create new expressions.
+ *
+ * @author Konstantin Kudryashov <ever.zet@gmail.com>
+ */
+class PartialNamedIncludingHiddenSelector extends PartialNamedSelector
```
aik099: How you plan to use this selector, e.g. during scenario writing for Behat?

src/Selector/PartialNamedIncludingHiddenSelector.php
claudiu-cristea: `PartialNamedIncludingHiddenFieldsSelector` would be better?

aik099: No, I've meant `notFieldTypeFilter` and `buttonTypeFilter` doesn't mention hidden fields in any way.

src/Selector/PartialNamedIncludingHiddenSelector.php
claudiu-cristea: Well, we need this in Drupal testing framework, see https://www.drupal.org/node/2767655. As you can see there, we init the session by adding also this new selector:

```
    $selectors_handler = new SelectorsHandler([
      'named_partial_including_hidden' => new PartialNamedIncludingHiddenSelector()
    ]);
    $session = new Session($driver, $selectors_handler);
```

Then we have an extension of WebAssert and override fieldExists() and fieldNotExists() to use also this new selector.

I was not thinking how this can be used in Behat scenarios. Probably that could be a follow up task?

aik099:
@stof , what do you think about this PR? Searching for hidden fields doesn't look like something, that will fit into Mink purpose.

src/Selector/PartialNamedIncludingHiddenSelector.php
claudiu-cristea: Wel, I cannot rename that because, in fact, I'm overriding there the existing `%notFieldTypeFilter%` replacement.

claudiu-cristea:
I think we should even consider this as the default behaviour for "named_partial". I can provide a PR if you agree.

dawehner:
A hidden field makes sense when you use a JS based driver, for example with https://github.com/jcalderonzumba/MinkPhantomJSDriver but I agree, per default, a browser user doesn't have the capability directly to do that.

On the other hand, making it possible to test security related stuff, by also changing the hidden fields, could be made as a point.

PR #716, opened 2016-09-26
Modifiying header test due to limitations found on PhantomJS

jcalderonzumba:
The purpose of this changes relates to the way headers test is being done.

PhantomJS only way to get the page HTML is via http://phantomjs.org/api/webpage/property/frame-content.html method and this is a know thing of phantomjs, it DOES changes the original HTML therefore in phantomjs drivers (https://github.com/jcalderonzumba/MinkPhantomJSDriver) test fails because the array "=>" changes to "=&gt".

Also in my personal opinion the test was a bit basic doing just a "dump" of $_SERVER global variable using print_r. In this PR it is changed to generate a "proper html" adding a 
`<p>/<p>` for each item of $_SERVER.

This is also part of the process to make this driver an official driver of Mink.

### DIFF
--- a/driver-testsuite/tests/Basic/HeaderTest.php
+++ b/driver-testsuite/tests/Basic/HeaderTest.php
@@ -25,7 +25,7 @@ class HeaderTest extends TestCase
         $this->getSession()->setRequestHeader('Accept-Language', 'fr');
         $this->getSession()->visit($this->pathTo('/headers.php'));
 
-        $this->assertContains('[HTTP_ACCEPT_LANGUAGE] => fr', $this->getSession()->getPage()->getContent());
+        $this->assertContains('HTTP_ACCEPT_LANGUAGE=fr', $this->getSession()->getPage()->getContent());
     }
 
     public function testSetUserAgent()
@@ -34,7 +34,7 @@ class HeaderTest extends TestCase
 
         $session->setRequestHeader('user-agent', 'foo bar');
         $session->visit($this->pathTo('/headers.php'));
-        $this->assertContains('[HTTP_USER_AGENT] => foo bar', $session->getPage()->getContent());
+        $this->assertContains('HTTP_USER_AGENT=foo bar', $session->getPage()->getContent());
     }
 
     public function testResetHeaders()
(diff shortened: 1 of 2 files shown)


aik099:
You should also send PR to browserkit driver repo, because it overrides file for printing headers.

jcalderonzumba:
Understood, will create PR accordingly.

jcalderonzumba:
@aik099 i think https://github.com/minkphp/MinkBrowserKitDriver/pull/113 that PR takes care of my changes.

aik099:
It does not. It moves fixtures to central place. It doesn't replace `[header name] => [header value]` output format into `header name=header value` as you did in this PR.

Moreover now your PR becomes invalid, since `driver-testsuite` folder was moved away from Mink repo into newly created https://github.com/minkphp/driver-testsuite repo, where you should send new PR, that combined both changes and close this PR.

stof:
I plan to refactor the way we print info in the driver testsuite, because we currently don't escape HTML fully.

aik099:
> I plan to refactor the way we print info in the driver testsuite, because we currently don't escape HTML fully.

We don't? I though changes addressing XSS issue that were made to test suite (the `htmlspecialchars` used in lots of places) already escape what's needed.

stof:
@aik099 XSS was address, but escaping before ``printf`` still outputs unescaped ``>`` in the HTML, which is previsely what causes issues for PhantomJS here (and probably also for recent versions of Selenium)

jcalderonzumba:
Sadly on PhantomJS we can not get the "raw" data sent from the server and the HTML you get is the interpreted HTML. 
Is not only about htmlentities but also about other things related to HTML parsing and rendering, i can give another example.
If your server sends:
`<div><p>Hi</p>` phantomjs will do it's best and will tell you that the HTML is
`<div><p>Hi</p></div>` therefore sometimes getContent will not be the same as you sent specially on bad formed html or in this particular case the ">".

jcalderonzumba:
@stof @aik099 Should i test now PhantomJSDriver against new driver test suite package and if needed ask the PR there?

aik099:
Yes, you should. All class names in test suite are the same. You just need to use different package to get tests and make a change in `phpunit.xml`.

jcalderonzumba:
@aik099 thanks, @stof already did the change for PhantomJSDriver and is merged, now what i got to do is send a PR to the new driver test suite package.

jcalderonzumba:
I think this PR does not make sense here so i'm closing and opening in the driver-testsuite repo.

stof:
@jcalderonzumba I have a branch locally starting to deal with this to stop outputting invalid HTML (Selenium has issues with it too now)

jcalderonzumba:
@stof that's good 👍

PR #718, opened 2016-09-28
Test only for #702: Cannot fill in fields with pipes (|) in name-attribute

claudiu-cristea:
Here's the test showing the bug. for #702

### DIFF
--- a/src/Selector/Xpath/Manipulator.php
+++ b/src/Selector/Xpath/Manipulator.php
@@ -24,6 +24,11 @@ class Manipulator
     const UNION_PATTERN = '/\|(?![^\[]*\])/';
 
     /**
+     * Pipe ('|') inside xpath strings placeholder.
+     */
+    const PIPE_PLACEHOLDER = '****PIPE-CHAR-PLACEHOLDER****';
+
+    /**
      * Prepends the XPath prefix to the given XPath.
      *
      * The returned XPath will match elements matching the XPath inside an element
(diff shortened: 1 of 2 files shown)


src/Selector/Xpath/Manipulator.php
```diff
@@ -43,6 +48,21 @@ public function prepend($xpath, $prefix)
             $prefix = '('.$prefix.')';
         }
 
+        // Get all xpath strings.
```
dawehner: We should add a documentation why we are doing the replacement here: Support "|" in actual strings

claudiu-cristea:
@dawehner I added the docs in the last commit. Thank you.

@stof I see a Scrutinizer failure but I cannot explain it. The details there doesn't help to much. Can you point me to what's wrong with my coding standards? Thank you.

claudiu-cristea:
I fixed one suggestion from Scrutinizer but I still have no ideea why Scrutinizer fails. Any idea?

stof:
Well, looks like the scrutinizer config changed, and so it detected more issue sin the existing codebase. Ignore this failure for now

claudiu-cristea:
@stof, thank you. Then I'm done except the solution is not considered acceptable. I agree that placeholdering the string could not be too elegant, but right now I have no other idea. I don't think we can find a decent `\Behat\Mink\Selector\Xpath\Manipulator::UNION_PATTERN` pattern to sove the issue at once.

stof:
The issue with placeholdering, is that this can break if the placeholder itself appears in the string.

claudiu-cristea:
@stof, that's correct. Now, the placeholder is safe regardless of what the xpath string contains.

src/Selector/Xpath/Manipulator.php
```diff
@@ -43,6 +48,28 @@ public function prepend($xpath, $prefix)
             $prefix = '('.$prefix.')';
         }
 
+        // Compute a safe pipe placeholder, not included in the xpath string.
+        $suffix = 0;
+        do {
+            $placeholder = self::PIPE_PLACEHOLDER_TEMPLATE.$suffix++;
+        } while (false !== strpos($xpath, $placeholder));
```
klausi: This should have a test case where "****PIPE-CHAR-PLACEHOLDER-0" is in the string to make sure raising the number here works.

klausi:
I think we should properly parse the XPath expression, opened #720

klausi:
I think the test case here is wrong. The pipe operator is withing braces [], so the XPath expression within the brace should NOT be prefixed. That's why the test case expectation is a bit different in my PR. As you can see from the other testdata and Manipulator::UNION_PATTERN the union operator should be ignored within braces.

stof:
Closing in favor of #720

PR #720, opened 2016-10-03
Allow pipes (|) in attributes by implementing better XPath parsing

klausi:
For issue #702 .

The pipe operator can appear in many places - a regex is too primitive, we need to actually parse the XPath expression.

This also uncovered a bug in the XPath expressions used for testing which contain a double single quote, which would be a broken XPath expression.

### DIFF
--- a/src/Selector/Xpath/Manipulator.php
+++ b/src/Selector/Xpath/Manipulator.php
@@ -44,7 +44,7 @@ class Manipulator
         }
 
         // Split any unions into individual expressions.
-        foreach (preg_split(self::UNION_PATTERN, $xpath) as $expression) {
+        foreach ($this->splitUnionParts($xpath) as $expression) {
             $expression = trim($expression);
             $parenthesis = '';
 
@@ -66,4 +66,50 @@ class Manipulator
 
         return implode(' | ', $expressions);
     }
+
+    /**
+     * Splits the XPath into parts that are separated by the union operator.
+     *
+     * @param string $xpath
+     *
+     * @return string[]
+     */
+    protected function splitUnionParts($xpath)
+    {
+        // Split any unions into individual expressions. We need to iterate
+        // through the string to correctly parse opening/closing quotes and
+        // braces which is not possible with regular expressions.
+        $union_parts = array();
+        $singleQuoteOpen = false;
+        $doubleQuoteOpen = false;
+        $bracketOpen = 0;
+        $lastUnion = 0;
+        for ($i = 0; $i < strlen($xpath); $i++) {
+            if ($xpath{$i} === "'" && $doubleQuoteOpen === false) {
+                if ($singleQuoteOpen === true) {
+                    $singleQuoteOpen = false;
+                } else {
+                    $singleQuoteOpen = true;
+                }
+            } else if ($xpath{$i} === '"' && $singleQuoteOpen === false) {
+                if ($doubleQuoteOpen === true) {
+                    $doubleQuoteOpen = false;
+                } else {
+                    $doubleQuoteOpen = true;
+                }
+            } else if ($singleQuoteOpen === false && $doubleQuoteOpen === false) {
+                if ($xpath{$i} === '[') {
+                    $bracketOpen++;
+                } else if ($xpath{$i} === ']') {
+                    $bracketOpen--;
+                } else if ($xpath{$i} === '|' && $bracketOpen === 0) {
+                    $union_parts[] = substr($xpath, $lastUnion, $i - $lastUnion);
+                    $lastUnion = $i + 1;
+                }
+            }
+        }
+        $union_parts[] = substr($xpath, $lastUnion);
+        return $union_parts;
+    }
+
 }
--- a/src/Selector/Xpath/Manipulator.php
+++ b/src/Selector/Xpath/Manipulator.php
@@ -74,12 +74,12 @@ class Manipulator
      *
      * @return string[]
      */
-    protected function splitUnionParts($xpath)
+    private function splitUnionParts($xpath)
     {
         // Split any unions into individual expressions. We need to iterate
         // through the string to correctly parse opening/closing quotes and
         // braces which is not possible with regular expressions.
-        $union_parts = array();
+        $unionParts = array();
         $singleQuoteOpen = false;
         $doubleQuoteOpen = false;
         $bracketOpen = 0;
@@ -103,13 +103,13 @@ class Manipulator
                 } else if ($xpath{$i} === ']') {
                     $bracketOpen--;
                 } else if ($xpath{$i} === '|' && $bracketOpen === 0) {
-                    $union_parts[] = substr($xpath, $lastUnion, $i - $lastUnion);
+                    $unionParts[] = substr($xpath, $lastUnion, $i - $lastUnion);
                     $lastUnion = $i + 1;
                 }
             }
         }
-        $union_parts[] = substr($xpath, $lastUnion);
-        return $union_parts;
+        $unionParts[] = substr($xpath, $lastUnion);
+        return $unionParts;
     }
 
 }
--- a/src/Selector/Xpath/Manipulator.php
+++ b/src/Selector/Xpath/Manipulator.php
@@ -87,11 +87,11 @@ class Manipulator
         for ($i = 0; $i < strlen($xpath); $i++) {
             $char = $xpath[$i];
 
-            if ($char === "'" && $inDoubleQuotedString === false) {
+            if ($char === "'" && !$inDoubleQuotedString) {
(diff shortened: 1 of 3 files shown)


src/Selector/Xpath/Manipulator.php
```diff
@@ -44,7 +44,7 @@ public function prepend($xpath, $prefix)
         }
 
         // Split any unions into individual expressions.
-        foreach (preg_split(self::UNION_PATTERN, $xpath) as $expression) {
```
claudiu-cristea: `self::UNION_PATTERN` constant should be removed too. Or is used in other places?

klausi: it is used further above and for backwards compatibility reasons we should not remove it, other libraries might rely on it.

klausi:
How do I see what failed on Scrutinizer? It just tells me that one failure condition exists, but no details?

src/Selector/Xpath/Manipulator.php
```diff
@@ -66,4 +66,50 @@ public function prepend($xpath, $prefix)
...
     }
+
+    /**
+     * Splits the XPath into parts that are separated by the union operator.
+     *
+     * @param string $xpath
+     *
+     * @return string[]
+     */
+    protected function splitUnionParts($xpath)
```
stof: should be private

src/Selector/Xpath/Manipulator.php
```diff
@@ -66,4 +66,50 @@ public function prepend($xpath, $prefix)
...
+    {
+        // Split any unions into individual expressions. We need to iterate
+        // through the string to correctly parse opening/closing quotes and
+        // braces which is not possible with regular expressions.
+        $union_parts = array();
```
stof: please use camelCase

tests/Selector/Xpath/ManipulatorTest.php
```diff
@@ -56,8 +56,13 @@ public function getPrependedXpath()
...
                 'some_xpath',
-                "some_tag1 | some_tag2[@foo =\n 'bar|'']\n | some_tag3[foo | bar]",
-                "some_xpath/some_tag1 | some_xpath/some_tag2[@foo =\n 'bar|''] | some_xpath/some_tag3[foo | bar]",
+                "some_tag1 | some_tag2[@foo =\n 'bar|']\n | some_tag3[foo | bar]",
+                "some_xpath/some_tag1 | some_xpath/some_tag2[@foo =\n 'bar|'] | some_xpath/some_tag3[foo | bar]",
+            ),
+            'containing pipe' => array(
+                'some_xpath',
+                "some_tag[(contains(normalize-space(string(.)), 'foo|bar') | other_tag[contains(./@some_attribute, 'foo|bar')])]",
+                "some_xpath/some_tag[(contains(normalize-space(string(.)), 'foo|bar') | other_tag[contains(./@some_attribute, 'foo|bar')])]",
```
stof: wrong expectation. Should be ``some_xpath/other_tag`` for the second part of the union

klausi: well, that's the question: the original regex suggests that within square brackets expressions should not be prefixed. Are you saying that we should also change the 'complex condition' test case further above to produce a different result?

src/Selector/Xpath/Manipulator.php
klausi: fixed

src/Selector/Xpath/Manipulator.php
klausi: fixed

klausi:
While we wait for feedback here I'm now integrating my Mink fork in our PHPUnit initiative development branch for Drupal https://github.com/klausi/drupal/commit/9db23a28bfb26ca1bf82e95e5fd6d157c182f912

tests/Selector/Xpath/ManipulatorTest.php
stof: ah sorry, I missed the fact it was inside square braces

src/Selector/Xpath/Manipulator.php
```diff
@@ -66,4 +66,50 @@ public function prepend($xpath, $prefix)
...
+     * @return string[]
+     */
+    private function splitUnionParts($xpath)
+    {
+        // Split any unions into individual expressions. We need to iterate
+        // through the string to correctly parse opening/closing quotes and
+        // braces which is not possible with regular expressions.
+        $unionParts = array();
+        $singleQuoteOpen = false;
+        $doubleQuoteOpen = false;
```
stof: I would name them ``$inSingleQuotedString`` and ``$inDoubleQuotedString``. It would better reflect their usage

src/Selector/Xpath/Manipulator.php
```diff
@@ -66,4 +66,50 @@ public function prepend($xpath, $prefix)
...
+        $bracketOpen = 0;
+        $lastUnion = 0;
+        for ($i = 0; $i < strlen($xpath); $i++) {
+            if ($xpath{$i} === "'" && $doubleQuoteOpen === false) {
```
stof: Please use ``!`` instead of ``=== false``

src/Selector/Xpath/Manipulator.php
```diff
@@ -66,4 +66,50 @@ public function prepend($xpath, $prefix)
...
+                if ($singleQuoteOpen === true) {
+                    $singleQuoteOpen = false;
```
stof: can be simplified: ``$singleQuoteOpen = !$singleQuoteOpen``

src/Selector/Xpath/Manipulator.php
```diff
@@ -66,4 +66,50 @@ public function prepend($xpath, $prefix)
...
+                } else {
+                    $singleQuoteOpen = true;
+                }
+            } else if ($xpath{$i} === '"' && $singleQuoteOpen === false) {
```
stof: should be ``elseif`` without space to match our coding standards

src/Selector/Xpath/Manipulator.php
```diff
@@ -66,4 +66,50 @@ public function prepend($xpath, $prefix)
...
+                if ($doubleQuoteOpen === true) {
+                    $doubleQuoteOpen = false;
+                } else {
+                    $doubleQuoteOpen = true;
+                }
+            } else if ($singleQuoteOpen === false && $doubleQuoteOpen === false) {
+                if ($xpath{$i} === '[') {
```
stof: please use ``$xpath[$i]`` rather than ``$xpath{$i}`` to access the element.

And I think defining a ``$char`` variable may be more readable.

src/Selector/Xpath/Manipulator.php
```diff
@@ -66,4 +66,50 @@ public function prepend($xpath, $prefix)
...
+     */
+    private function splitUnionParts($xpath)
+    {
+        // Split any unions into individual expressions. We need to iterate
+        // through the string to correctly parse opening/closing quotes and
+        // braces which is not possible with regular expressions.
+        $unionParts = array();
+        $singleQuoteOpen = false;
+        $doubleQuoteOpen = false;
+        $bracketOpen = 0;
```
stof: ``$bracketLevel`` or ``$openedBrackets`` would be a better name

src/Selector/Xpath/Manipulator.php
klausi: fixed

src/Selector/Xpath/Manipulator.php
klausi: fixed

src/Selector/Xpath/Manipulator.php
klausi: fixed

src/Selector/Xpath/Manipulator.php
klausi: fixed

src/Selector/Xpath/Manipulator.php
klausi: fixed

src/Selector/Xpath/Manipulator.php
klausi: fixed

claudiu-cristea:
@klausi, looks good, better than trying with regexp. That seems to be an impossible try. When this get committed we need only to refresh composer.lock in https://www.drupal.org/node/2808085

stof:
thanks @klausi

stof:
FYI, the discussion on the related Symfony BrowserKit PR highlighted the high performance impact of this change. So I'm looking at the progress on the alternate implemented proposed by @nicolas-grekas in https://github.com/symfony/symfony/pull/20235 before making a Mink release (to use it instead if appropriate)

PR #721, opened 2016-10-14
Remove the driver testsuite

### LINKED ISSUE #717: [RFC] Extract the driver testsuite to a dedicated package
Currently, the driver testsuite is maintained in the main Mik repo. but this leads to some weirdness:

- changes in the driver testsuite are done in the Mink repo, but do no impact the Mink CI (they are only about drivers)
- as we want the drivers to access it to run their tests, we cannot exclude it from the github archive, leading to the driver testsuite being installed in prod (hopefully not web accessible, but this is what triggers #689)
- drivers have to do weird things to be able to autoload classes in the driver testsuite (actually, this weird thing got moved to the driver testsuite itself, but then drivers have to use a PHPUnit bootstrap file from the vendor, which is also weird), as Mink is a dependency there and so ``autoload-dev`` cannot help us
- testing locally would be much easier if the driver testsuite could provide a script launching the necessary PHP webserver for fixtures (today, people must either find the right command manually, for instance by looking at our travis config, or configure a virtual host in their webserver), which could also remove the mandatory need for a webfixtures host (the phpunit.xml.dist in each driver could provide a value corresponding to the default value of this script instead of forcing you to provide a value). This is hard to do today, while a separate package could expose it as a composer bin script.

Most driver testsuite changes do no include any changes to Mink itself in the same PR. So the need for synchronisation of merges between Mink and the driver testsuite would be quite small, as this happens only in 1 case: when breaking BC in Mink (which should not happen outside of major versions). If we want to test a new feature, sync is not really needed. We just need to merge in Mink *before* merging in the driver testsuite (and bumping the driver testsuite min requirement for Mink in the PR using the new feature, or skipping the test).

So I see mostly advantages.
Note that this is also the approach used for the PHPCR testsuite (there is a package dedicated to it, and each implementation has a dev requirement on it).

what do you think @aik099 ?

stof:
The driver testsuite has been moved to https://github.com/minkphp/driver-testsuite.

Closes #717


stof:
It's gone... :boom:

PR #722, opened 2016-10-18
Optimize the splitting of XPath unions

stof:
Consuming each char of the XPath string one by one is slower than necessary, as only a few chars are interesting for us.
This implementation written by @nicolas-grekas for Symfony BrowserKit in https://github.com/symfony/symfony/pull/20235 (and slightly adapted for our needs) is much faster by consuming irrelevant chars in batch.

See https://gist.github.com/stof/5131c79c576600800aac7313b182d44d for the benchmark:
- the (broken) Mink 1.7 implementation is the `regex` one
- the implementation merged in #720 is the `iteration`
- the implementation submitted in this PR is the `grekas_checking` one

Note that I added additional comments compared to the code written by Nicolas to explain what actually happens.

### DIFF
--- a/src/Selector/Xpath/Manipulator.php
+++ b/src/Selector/Xpath/Manipulator.php
@@ -76,56 +76,54 @@ class Manipulator
      */
     private function splitUnionParts($xpath)
     {
-        // Split any unions into individual expressions. We need to iterate
-        // through the string to correctly parse opening/closing quotes and
-        // braces which is not possible with regular expressions.
-        $unionParts = array();
-        $inSingleQuotedString = false;
-        $inDoubleQuotedString = false;
-        $openedBrackets = 0;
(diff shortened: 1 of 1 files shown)


stof:
@klausi sorry that your first contribution to Mink will mostly be replaced after a few days only. I hope your next contribution will last longer...

aik099:
Looking at the code alone should be quite performant. Luckily we have unit tests in place to confirm, that xpath processing wasn't broken as a result.

stof:
@aik099 for performance, you can also look at the benchmark (the gist also contains results if you scroll)

klausi:
@stof no worries, I'm just happy that we got it all fixed so quickly.

I can confirm that the change in this PR does not break in Drupal and behaves the same.

Issue #714, opened 2016-08-26
fieldset named selector returns the most outer fieldset instead of the closest ancestor


cyberwolf:
In an attempt to use the fieldset named selector, it did not return the fieldset I expected it to return when using find() with the legend value as a locator and fieldsets are nested. Instead of the closest ancestor fieldset a legend belongs to, it returns the farthest ancestor fieldset.
An example:
* fieldset 1
** fieldset 1.1
*** legend "a"
** fieldset 1.2
*** legend "b"

When using the fieldset named selector with the value of either legend "a" or legend "b", in both cases fieldset 1 is returned, while I expected it to return respectively fieldset 1.1 and fieldset 1.2

aik099:
It would be better demonstrated with a failing test case sent as PR or at least HTML and PHP code posted in here to demonstrate the problem.

cyberwolf:
@aik099 sure, I'll provide a test.

claudiu-cristea:
Should be closed in favour of #726.

PR #726, opened 2016-11-02
Tryin' to select the inner fieldset, is selecting both

claudiu-cristea:
The test proving the bug.


claudiu-cristea:
I think is the same bug as in #714

stof:
The issue is that the outer fieldset also contains a legend containing the requested text. The current XPath does not enforce that there is no other fieldset between them (I don't know whether enforcing it is easy in XPath)

PR #729, opened 2017-02-02
:arrow_up: Remove PHP 5.3 from Travis CI test matrix

alexislefebvre:
PHP 5.3 is not supported since 2014 : http://php.net/supported-versions.php


alexislefebvre:
We can also remove PHP 5.4 and 5.5 that have reached end of life: http://php.net/eol.php

CHANGES.md
```diff
@@ -1,3 +1,10 @@
+1.7.2 / 2017-MM-DD
```
alexislefebvre: I took this idea from PHPUnit's changelog: https://github.com/sebastianbergmann/phpunit/commit/69f832b87c731d5cacad7f91948778fe98335fdd

stof:
I would not remove them all yet. We still advocate PHP 5.3+ support. And projects using Mink in their testsuite might need to keep running their tests on these PHP version (for instance, Drupal would hate us if we drop PHP 5.5).

And dropping support does not simplify our work at all anyway

alexislefebvre:
@stof Thanks, I understand that we can't break projects that depend on Mink. How long do you plan to support PHP 5.3? Should I close this PR for the moment?

aik099:
Supporting PHP 5.3 or any other PHP version doesn't really hurt Mink development team/community at the moment. When, and only when, it will then we'll reconsider supported PHP versions.

I believe you can close PR now. Thank you for your concerns.

Issue #731, opened 2017-02-06
Auto-start session only upon 1st "->visit(...)" method call breaks my test suite


peterrehm:
PR #705 from @aik099 breaks my testsuite as I am resizing the Window prior to each scenario.

````php
class FeatureContext extends RawMinkContext
{
    use KernelDictionary;

    /**
     * @BeforeStep
     */
    public function beforeScenario()
    {
        if ($this->getSession()->getDriver() instanceof Selenium2Driver) {
            $this->getSession()->resizeWindow(1440, 900, 'current');
        }
    }
}
````

This breaks with the error:

````
    ┌─ @BeforeScenario # AppBundle\Features\Context\FeatureContext::beforeScenario()
    │
    ╳  Fatal error: Call to a member function window() on null (Behat\Testwork\Call\Exception\FatalThrowableError)
    │
````

aik099:
That is unexpected.

Any other methods, except `resizeWindow` that can use session before `->visit(...)` is called? We can auto-start session in them as well. I hope there are no much such methods.

peterrehm:
I am not using others, the question is if there are other. As we cannot know what others want to do prior the forst visit call I would suggest then leave it as it is but maybe optimize the session start prozedure.

Right now if you call session start with the selenium driver twice you get different sessions. Only if I add the check if the session is started and start it conditionally it works as expected. 

How about throwing an exception when you call start on an already started session or only start if it has not been started previously as you are doing here: https://github.com/minkphp/Mink/pull/705/files#diff-06f6ce27521fd9b588fa1d23c8ad02d1R143

aik099:
> I am not using others, the question is if there are other. ...

Here are the list of methods (your's included), that might also need to auto-start session:

* `resizeWindow`
* `maximizeWindow`

> How about throwing an exception when you call start on an already started session or only start if it has not been started previously

We can do that probably. Also when attempting to stop non-started session the exception can be thrown. Nobody really experienced issues with that before by the way.

stof:
> Right now if you call session start with the selenium driver twice you get different sessions.

I would consider it a bug in the selenium driver. It should not start a second time if it is already started

stof:
or we can indeed handle it at the Session level, solving it for all drivers

peterrehm:
Shall I submit a PR for the autostart and the session handling?
So just silently do not restart it if already started?

stof:
yeah, please submit a PR

aik099:
[...]

PR #732, opened 2017-02-07
Only start session if the session has not been started before

peterrehm:
As discussed in #731.

### DIFF
--- a/src/Session.php
+++ b/src/Session.php
@@ -59,8 +59,6 @@ class Session
      *
      * Calling any action before visiting a page is an undefined behavior.
      * The only supported method calls on a fresh driver are
-     * - resizeWindow()
-     * - maximizeWindow()
      * - visit()
      * - setRequestHeader()
      * - setBasicAuth()
@@ -96,8 +94,6 @@ class Session
      *
      * Calling any action before visiting a page is an undefined behavior.
      * The only supported method calls on a fresh driver are
-     * - resizeWindow()
-     * - maximizeWindow()
      * - visit()
      * - setRequestHeader()
      * - setBasicAuth()
--- a/src/Session.php
+++ b/src/Session.php
@@ -55,7 +55,7 @@ class Session
     }
 
     /**
-     * Starts session driver.
+     * Starts session driver if session is not started.
      *
      * Calling any action before visiting a page is an undefined behavior.
      * The only supported method calls on a fresh driver are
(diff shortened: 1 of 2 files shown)


tests/SessionTest.php
```diff
@@ -83,34 +101,10 @@ public function testRestart()
         $this->session->restart();
     }
 
-    public function testVisitWithoutRunningSession()
```
stof: this test must not be removed. We still need to cover the case of calling ``visit`` on a started and on a non-started session IMO (thus, your mocks don't describe the actual interaction happening, as you removed the mocking of ``isStarted``

src/Session.php
```diff
@@ -67,7 +67,10 @@ public function isStarted()
      */
     public function start()
     {
-        $this->driver->start();
+        // start session if needed
```
stof: useless comment. The phpdoc already says it

src/Session.php
```diff
@@ -55,10 +55,12 @@ public function isStarted()
     }
 
     /**
-     * Starts session driver.
+     * Starts session driver if session is not started.
      *
      * Calling any action before visiting a page is an undefined behavior.
      * The only supported method calls on a fresh driver are
+     * - resizeWindow()
+     * - maximizeWindow()
```
stof: if we want to allow this, we must add new tests in the driver testsuite calling these methods before visiting a page, to ensure it actually works in all drivers

stof: btw, this does not seem related to this PR, but to #733

peterrehm: Correct, that is my bad. Where would you add the tests? Looking at the CoreDriverTest and the Tests in the Selenium driver I could not see any tests on e.g. the visit method

peterrehm: I have moved that to #733.

stof: the driver testsuite lives in https://github.com/minkphp/driver-testsuite and is a dev dependency in all drivers repositories, to share the same testsuite in all of them.

peterrehm: That makes sense. I was scanning the testsuite, I could not find a test where the autostart is tested e.g. for the visit method by grepping te repository. Do you know if there is anything in this regard?

stof: Allowing to call these methods without visiting a page is **not** related to auto-starting. It is a different feature (and this is what you need to test)

peterrehm: Oh okay. So what do you think about: https://github.com/minkphp/driver-testsuite/pull/10

tests/SessionTest.php
aik099: Actually the discussion from https://github.com/minkphp/Mink/pull/733/files#r99773293 is meant to happen on this PR instead.

PR #733, opened 2017-02-07
Session autostart

### LINKED ISSUE #731: Auto-start session only upon 1st "->visit(...)" method call breaks my test suite

peterrehm:
As discussed in #731.

Will be rebased after #732 is merged.

Tickets:
- fixes #731

### DIFF
--- a/src/Session.php
+++ b/src/Session.php
@@ -59,8 +59,6 @@ class Session
      *
      * Calling any action before visiting a page is an undefined behavior.
      * The only supported method calls on a fresh driver are
-     * - resizeWindow()
-     * - maximizeWindow()
      * - visit()
      * - setRequestHeader()
      * - setBasicAuth()
@@ -96,8 +94,6 @@ class Session
      *
      * Calling any action before visiting a page is an undefined behavior.
      * The only supported method calls on a fresh driver are
-     * - resizeWindow()
-     * - maximizeWindow()
      * - visit()
      * - setRequestHeader()
      * - setBasicAuth()
--- a/src/Session.php
+++ b/src/Session.php
@@ -363,6 +363,8 @@ class Session
      */
     public function resizeWindow($width, $height, $name = null)
     {
+        // start session if needed
+        $this->start();
         $this->driver->resizeWindow($width, $height, $name);
     }
 
@@ -373,6 +375,8 @@ class Session
      */
     public function maximizeWindow($name = null)
     {
+        // start session if needed
+        $this->start();
         $this->driver->maximizeWindow($name);
     }
 }
--- a/src/Session.php
+++ b/src/Session.php
@@ -55,7 +55,7 @@ class Session
     }
 
     /**
-     * Starts session driver.
+     * Starts session driver if session is not started.
      *
      * Calling any action before visiting a page is an undefined behavior.
      * The only supported method calls on a fresh driver are
@@ -67,7 +67,10 @@ class Session
      */
     public function start()
     {
-        $this->driver->start();
+        // start session if needed
+        if (!$this->isStarted()) {
+            $this->driver->start();
+        }
     }
 
     /**
@@ -141,9 +144,7 @@ class Session
     public function visit($url)
     {
         // start session if needed
-        if (!$this->isStarted()) {
-            $this->start();
-        }
+        $this->start();
 
         $this->driver->visit($url);
     }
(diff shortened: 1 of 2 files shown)


src/Session.php
```diff
@@ -141,9 +144,7 @@ public function getSelectorsHandler()
     public function visit($url)
     {
         // start session if needed
```
aik099: You can remove `// start session if needed` comment from all 3 methods, because it's already present in `start` method.

peterrehm:
Updated

src/Session.php
```diff
@@ -67,7 +67,10 @@ public function isStarted()
      */
     public function start()
     {
-        $this->driver->start();
+        // start session if needed
```
stof: this comment is useless. The code is clear enough (and the phpdoc already says it too)

src/Session.php
```diff
@@ -363,6 +362,7 @@ public function wait($time, $condition = 'false')
      */
     public function resizeWindow($width, $height, $name = null)
     {
+        $this->start();
```
stof: technically, these actions are not defined as supported before visiting (see the phpdoc of ``start``)

tests/SessionTest.php
```diff
@@ -83,34 +101,10 @@ public function testRestart()
         $this->session->restart();
     }
 
-    public function testVisitWithoutRunningSession()
```
stof: this test must not be removed

src/Session.php
aik099: But @peterrehm is using them and they work perfectly. Otherwise there is no way to open window initially with given size so that upon page visiting we won't be resizing it again.

tests/SessionTest.php
peterrehm: I removed the test as I added the tests for the start() method.
As start() is tested for both options I thought it is enough to test just that start() 
is called in e.g. the visit method.

src/Session.php
peterrehm: Removed in the other PR.

src/Session.php
peterrehm: Updated the list of Methods i the phpdoc. Is this what you meant?

tests/SessionTest.php
stof: well, the test you kept describes the wrong interaction with the driver (it does not describe the usage of ``isStarted``.

and I would keep the test ensuring we can visit without starting, as the test does not need to know that the code is calling ``$this->start()`` internally rather than directly calling the driver.

peterrehm: But the test ensures that start is called:

````php
    public function testVisit()
    {
        $this->driver
            ->expects($this->once())
            ->method('start');

        $this->driver
            ->expects($this->once())
            ->method('visit')
            ->with($url = 'some_url');

        $this->session->visit($url);
    }
````

The test with isStarted is not contained in `testStartWithoutRunningSession` and `testStartWithRunningSession` as the logic is moved to the `start` method.

stof: But when you call ``$this->visit``, the ``isStarted`` method of the driver is still called. So your description of the interaction is incomplete (and btw, because it is incomplete, ``isStarted`` would return ``null``, not ``false``).

PHPUnit mocks don't complain about incomplete interaction specifications, which is why your test is passing (and a reason why I actually prefer Prophecy, but I haven't converted the whole Mink testsuite to it)

peterrehm: But IMHO it doesn't matter in this case as we are finde if started() is called and afterwards the visit() method is called. Otherwise tests would need to be duplicated for the other autostart methods as well. 

If you want I can add this in.

aik099: @stof , @stof , any updates? This seem to be last change to be made before merging.

P.S.
Change actually needs to be made on #732 instead and this PR rebased on top of it.

PR #738, opened 2017-05-18
Suggest chrome-mink-driver as fast JS-enabled alternative to selenium

DorianCMore:
This driver has all the benefits of selenium2 without the performance cost.

It supports features unsupported by selenium2driver:

- setBasicAuth
- setRequestHeader
- getResponseHeaders
- getStatusCode

And it only lacks one method that Selenium2Driver implements, namely 'switchToIFrame'.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -27,6 +27,7 @@
         "behat/mink-browserkit-driver": "extremely fast headless driver for Symfony\\Kernel-based apps (Sf2, Silex)",
         "behat/mink-goutte-driver":     "fast headless driver for any app without JS emulation",
         "behat/mink-selenium2-driver":  "slow, but JS-enabled driver for any app (requires Selenium2)",
+        "dmore/chrome-mink-driver":     "fast and JS-enabled driver for any app (requires chromium or google chrome)",
         "behat/mink-zombie-driver":     "fast and JS-enabled headless driver for any app (requires node.js)"
     },
 


aik099:
Do I need to start something (e.g. server) on computer to use this driver?

The documentation https://gitlab.com/DMore/chrome-mink-driver page suggests that it's a headless driver then I don't need Chrome installed/started for it to work?

stof:
I see several issues for your driver:

- the composer package has requirements on Behat and MinkExtension. This is not OK for a Mink driver (Mink is independent from Behat). If you also have a Behat extension adding support for your driver in MinkExtension, it should be a separate package (which would depend on your driver)
- the Mink requirement for you package is wrong. You require `1.*` but your class extends CoreDriver, which does not exist in the 1.0 release. As you even implement `findElementXpaths` rather than `find`, you min support version is 1.7. so the requirement should be `^1.7`.
- `findElementXpaths` should be protected, not public
- the documentation must explain how to run it (do you need to have Chrome running ?)

stof:
Btw, rather than adding composer suggestions for third-party drivers, I would prefer to create a dedicated section in the official documentation to talk about third-party drivers

DorianCMore:
Thanks for the through feedback. I have addressed all your points in 2.0, since removing the behat extension is a BC break. 

The behat extension is here: https://packagist.org/packages/dmore/behat-chrome-extension

stof:
@DorianCMore you forgot to change the requirements in 2.0, so it still depends on Behat and MinkExtension

DorianCMore:
@stof fixed now.

Regarding your documentation suggestion, I think most people won't RTFM and go with one of the suggested drivers. Which means either compromising on speed or features.

Which is a pity because they can get zombiejs performance with selenium features. https://twitter.com/DorianCMore/status/865109382291488768

If you'd be more comfortable forking it then go ahead. It's MIT licensed.

stof:
@DorianCMore I'm not convince they read the composer suggestions more than the documentation actually (probably the opposite).

DorianCMore:
They're shown on packagist and on composer install. They get more visibility without being the better options.

Obviously I'm biased, but IMO selenium should only be recommended for testing browsers other than chrome.

And zombie is nowhere close to "best of both worlds" now. Chrome saves you the pain of having to reproduce a zombie failure in a real browser, and even lets you see it with your own eyes by disabling the --headless flag, or taking screenshots. If anything, I would advise against ever using zombie nowadays.

peterrehm:
I think it would be good to merge this so people are aware of this drivers.

aik099:
Actually we have drivers listed in both documentation (http://mink.behat.org/en/latest/) and `composer.json` file. Based on that this PR can be merged, but we need connected PR for documentation.

jason69:
@aik099 Mmmhhh weird, I didn't find any mention of chrome-mink-driver on http://mink.behat.org/en/latest/

aik099:
@jason69 , that's why I told that we need another PR for documentation.

adorin:
Documentation PR submitted here: https://github.com/minkphp/docs/pull/82

aik099:
Merged, thanks @DorianCMore & @adorin .

PR #742, opened 2017-10-16
Document that evaluateScript() has specific needs

timplunkett:


src/Session.php
```diff
@@ -332,6 +332,10 @@ public function executeScript($script)
     /**
      * Execute JS in browser and return it's response.
      *
+     * The "return" keyword is optional in the script passed as argument.
```
dawehner: It would be nice to have an actual example.

stof: see https://github.com/minkphp/driver-testsuite/blob/master/tests/Js/JavascriptEvaluationTest.php for the tests

aik099:
@timplunkett you should also update https://github.com/minkphp/Mink/blob/master/src/Driver/DriverInterface.php#L576-L589

timplunkett:
I'm used to putting docs about the parameters on the line after the `@param`, indented an extra two spaces. Doesn't seem to be how it's done here, so I just slapped it up top. lmk if you'd prefer it elsewhere

PR #744, opened 2017-10-17
Add new function WebAssert::responseContainsCount

NickWilde1990:


NickWilde1990:
Could probably be done with WebAssert::elementsCount or a regexp (depending on the exact needs), but not as easily or as good of performance.

stof:
case sensitivity is not consistent with `responseContains`

NickWilde1990:
@stof Updated to be case insensitive, thanks.

PR #746, opened 2017-11-18
Trying to run on PHP7.1 [#SymfonyConHackday2017]

Dottyx:


.travis.yml
```diff
@@ -2,7 +2,7 @@ language: php
 
 sudo: false
 
-php: [5.3, 5.4, 5.5, 5.6, 7.0, hhvm]
+php: [5.3, 5.4, 5.5, 5.6, 7.1, hhvm]
```
aik099: Please return PHP 7.0 back and of course keep PHP 7.1.

---

I see no reasons for dropping PHP 7.0 support.

tests/Driver/CoreDriverTest.php
```diff
@@ -65,7 +65,7 @@ public function testInterfaceMethods(\ReflectionMethod $method)
 
         $driver = $this->getMockForAbstractClass('Behat\Mink\Driver\CoreDriver');
 
-        $this->setExpectedException('Behat\Mink\Exception\UnsupportedDriverActionException');
+        $this->expectException('Behat\Mink\Exception\UnsupportedDriverActionException');
```
aik099: 1. Where `expectException` method is defined, PHPUnit or Mink?
2. If that is PHPUnit, then are you sure, that method is present in every PHPUnit version where test suite will be used?

tests/Driver/CoreDriverTest.php
```diff
@@ -102,7 +102,7 @@ private function getArgument(\ReflectionParameter $argument)
         if ($argument->getClass()) {
             return $this->getMockBuilder($argument->getClass()->getName())
                 ->disableOriginalConstructor()
-                ->getMock();
+                ->createMock();
```
aik099: Is `createMock` present in every PHPUnit version where test suite will be used?

tests/bootstrap.php
```diff
@@ -0,0 +1,12 @@
+<?php
+include_once __DIR__.'/../vendor/autoload.php';
+
+$classLoader = new \Composer\Autoload\ClassLoader();
+$classLoader->addPsr4("Behat\\Mink\\Tests\\", __DIR__, true);
+$classLoader->register();
+
+// PHPUnit 6 introduced a breaking change that
```
aik099: To address this you need to:

1. across test suite use `\PHPUnit\Framework\TestCase` (because it was available alongside with `\PHPUnit_Framework_TestCase`)
2. don't create any bootstrap file

composer.json
```diff
@@ -16,11 +16,11 @@
 
     "require": {
         "php":                   ">=5.3.1",
-        "symfony/css-selector":  "~2.1|~3.0"
+        "symfony/css-selector":  "~2.1|~3.0|dev-master"
```
derrabus: Should be `"~2.1|~3.0|~4.0"`

PR #747, opened 2017-11-18
Symfony 4 support added

JonEastman:
Added Symfony 4 support added. 

#SymfonyConHackday2017

PR #749, opened 2017-11-24
Allow Symfony 4. Fix Travis build

dunglas:
Alternative to #747.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -29,4 +29,5 @@ install:
 script: vendor/bin/simple-phpunit -v --coverage-clover=coverage.clover
 
 after_script:
-  - if [[ "7.0" != "$TRAVIS_PHP_VERSION" && "$TRAVIS_PHP_VERSION" != "hhvm" ]]; then wget https://scrutinizer-ci.com/ocular.phar && php ocular.phar code-coverage:upload --format=php-clover coverage.clover; fi
+  - wget https://scrutinizer-ci.com/ocular.phar
+  - php ocular.phar code-coverage:upload --format=php-clover coverage.clover
--- a/.travis.yml
+++ b/.travis.yml
@@ -7,20 +7,24 @@ php: [5.4, 5.5, 5.6, 7.0, 7.1, hhvm]
 matrix:
   fast_finish: true
   include:
+    - php: 5.3
+      dist: precise
     - php: 7.1
       env: COMPOSER_FLAGS='--prefer-lowest --prefer-stable' SYMFONY_DEPRECATIONS_HELPER=weak
-    - php: 5.4
+    - php: 7.1
       env: DEPENDENCIES=dev
 
 cache:
   directories:
     - $HOME/.composer/cache/files
+    - $HOME/symfony-bridge/.phpunit
 
 before_install:
   - if [ "$DEPENDENCIES" = "dev" ]; then composer config minimum-stability dev; fi;
 
 install:
   - composer update $COMPOSER_FLAGS
+  - vendor/bin/simple-phpunit install
 
 script: vendor/bin/simple-phpunit -v --coverage-clover=coverage.clover
 
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,14 +2,14 @@ language: php
 
 sudo: false
 
-php: [5.3, 5.4, 5.5, 5.6, 7.0, hhvm]
+php: [5.4, 5.5, 5.6, 7.0, 7.1, hhvm]
 
 matrix:
   fast_finish: true
   include:
-    - php: 5.3
+    - php: 7.1
       env: COMPOSER_FLAGS='--prefer-lowest --prefer-stable' SYMFONY_DEPRECATIONS_HELPER=weak
-    - php: 5.6
+    - php: 5.4
       env: DEPENDENCIES=dev
 
 cache:
@@ -17,13 +17,12 @@ cache:
     - $HOME/.composer/cache/files
 
 before_install:
-  - composer self-update
-  - if [ "$DEPENDENCIES" = "dev" ]; then perl -pi -e 's/^}$/,"minimum-stability":"dev"}/' composer.json; fi;
+  - if [ "$DEPENDENCIES" = "dev" ]; then composer config minimum-stability dev; fi;
 
 install:
   - composer update $COMPOSER_FLAGS
 
-script: phpunit -v --coverage-clover=coverage.clover
+script: vendor/bin/simple-phpunit -v --coverage-clover=coverage.clover
 
 after_script:
   - if [[ "7.0" != "$TRAVIS_PHP_VERSION" && "$TRAVIS_PHP_VERSION" != "hhvm" ]]; then wget https://scrutinizer-ci.com/ocular.phar && php ocular.phar code-coverage:upload --format=php-clover coverage.clover; fi
--- a/composer.json
+++ b/composer.json
@@ -16,11 +16,11 @@
 
     "require": {
         "php":                   ">=5.3.1",
-        "symfony/css-selector":  "~2.1|~3.0"
+        "symfony/css-selector":  "^2.7|^3.0|^4.0"
     },
 
     "require-dev": {
-        "symfony/phpunit-bridge": "~2.7|~3.0"
+        "symfony/phpunit-bridge": "^3.3|^4.0"
     },
 
     "suggest": {


dunglas:
And Travis is happy!

.travis.yml
```diff
@@ -2,28 +2,27 @@ language: php
 
 sudo: false
 
-php: [5.3, 5.4, 5.5, 5.6, 7.0, hhvm]
+php: [5.4, 5.5, 5.6, 7.0, 7.1, hhvm]
```
stof: Removing CI on 5.3 without dropping support is a no-go. This is the best way to drop support without telling it to Composer.

.travis.yml
```diff
@@ -2,28 +2,27 @@ language: php
...
 
 matrix:
   fast_finish: true
   include:
-    - php: 5.3
+    - php: 7.1
       env: COMPOSER_FLAGS='--prefer-lowest --prefer-stable' SYMFONY_DEPRECATIONS_HELPER=weak
-    - php: 5.6
+    - php: 5.4
```
stof: Not good. This would not test against a dev version of SF for instance, because they dropped support for 5.4. This job should run on the highest stable PHP version instead

.travis.yml
```diff
@@ -2,28 +2,27 @@ language: php
...
   directories:
     - $HOME/.composer/cache/files
 
 before_install:
-  - composer self-update
-  - if [ "$DEPENDENCIES" = "dev" ]; then perl -pi -e 's/^}$/,"minimum-stability":"dev"}/' composer.json; fi;
+  - if [ "$DEPENDENCIES" = "dev" ]; then composer config minimum-stability dev; fi;
 
 install:
   - composer update $COMPOSER_FLAGS
```
stof: please also add `vendor/bin/simple-phpunit install` here, so that installation does not get mixed in the test output

.travis.yml
stof: There is a trick to force it to keep running on 5.3 precise for now.

.travis.yml
```diff
@@ -2,28 +2,27 @@ language: php
...
-    - php: 5.3
+    - php: 7.1
       env: COMPOSER_FLAGS='--prefer-lowest --prefer-stable' SYMFONY_DEPRECATIONS_HELPER=weak
-    - php: 5.6
+    - php: 5.4
       env: DEPENDENCIES=dev
 
 cache:
   directories:
     - $HOME/.composer/cache/files
```
stof: you should also ensure that the PHPUnit install performed by the bridge is cached whenever possible. See https://github.com/KnpLabs/KnpMenu/commit/170f8d7530d263116bd72b8047014494bc187a05 for an example

.travis.yml
dunglas: What do you prefer @stof? Using the trick for Travis (I know which one) or dropping the support?

aik099: To fix PHP 5.3 build you need to move it from general config to matrix, that uses `precise` OS version for that PHP version specifically.

dunglas: @aik099 yes I know ;).

dunglas:
@stof all comments fixed

sroze:
ping @stof :)

ruudk:
Any news on this? I feel like this is the last step to get Symfony2Extension compatible with SF 4 :pray:

sroze:
Alright, we can't block Symfony's Behat extension with that for that long. I'll remove Mink from the dependencies for now, do a release and maybe add it back at some point when this is merged.

.travis.yml
```diff
@@ -2,28 +2,32 @@ language: php
...
-  - composer self-update
-  - if [ "$DEPENDENCIES" = "dev" ]; then perl -pi -e 's/^}$/,"minimum-stability":"dev"}/' composer.json; fi;
+  - if [ "$DEPENDENCIES" = "dev" ]; then composer config minimum-stability dev; fi;
 
 install:
   - composer update $COMPOSER_FLAGS
+  - vendor/bin/simple-phpunit install
 
-script: phpunit -v --coverage-clover=coverage.clover
+script: vendor/bin/simple-phpunit -v --coverage-clover=coverage.clover
```
aik099: What is `simple-phpunit` and how it's better than regular phpunit?

dunglas: https://symfony.com/doc/current/components/phpunit_bridge.html

Here it is basically used to run tests both on PHP 5 and 7 (not easy with PHPUnit directly).

aik099:
> Alright, we can't block Symfony's Behat extension with that for that long. I'll remove Mink from the dependencies for now, do a release and maybe add it back at some point when this is merged.

@sroze , is Mink a dependency of Behat? I thought that only MinkExtension needs it. If you plan to release new version of MinkExtension without Mink listed as dependency, then it would be very confusing.

@dunglas , I wonder why we just can't add Symfony 4 in `composer.json` and that's it? This is the way how it was done for other projects. In that case PR could be merged right away (as long as all tests pass of course).

.travis.yml
sroze: for you @dunglas

sroze:
@aik099 it's a dev dependency, because of things like that:
https://github.com/Behat/Symfony2Extension/blob/c59756034c0837e11f36806893c308fe27fe9c4c/src/Behat/Symfony2Extension/ServiceContainer/Symfony2Extension.php#L51-L53

.travis.yml
aik099: Never had any issues with this. Just use PHPUnit version that Travis CI gives you (the code you've changed was doing that) and PHPUnit version compatible with used PHP version would be picked automatically.

aik099:
@sroze , understood. Now all we're waiting for @dunglas to fix issues reported by @stof .

sroze:
He fixed them already 😉 

https://github.com/minkphp/Mink/pull/749#issuecomment-346882884

.travis.yml
aik099: @dunglas , not fixed.

aik099:
@sroze , not the one where PHP 5.3 was removed from the build.

dunglas:
@aik099 it is actually fixed: https://travis-ci.org/minkphp/Mink/jobs/306878662

.travis.yml
```diff
@@ -2,28 +2,32 @@ language: php
...
 
 after_script:
-  - if [[ "7.0" != "$TRAVIS_PHP_VERSION" && "$TRAVIS_PHP_VERSION" != "hhvm" ]]; then wget https://scrutinizer-ci.com/ocular.phar && php ocular.phar code-coverage:upload --format=php-clover coverage.clover; fi
+  - wget https://scrutinizer-ci.com/ocular.phar
```
aik099: @dunglas , any idea why these 2 lines now can be performed for all PHP versions, but before for all but PHP7 and HHVM?

aik099:
> @aik099 it is actually fixed: https://travis-ci.org/minkphp/Mink/jobs/306878662

Ups. When discussion happens on one code line, but changes related to it are added to another code line it's hard to trace. That's fixed good.

Now only my last concern about need to use `simple-phpunit` (see relevant comment) and we're done.

.travis.yml
dunglas: Because now ocular supports PHP 7 (it wasn't the case when the initial PHP 7 support has been added to `.travis.yml` I guess.

dunglas:
@aik099 I replied to that, it's the only reliable way to run tests on both PHP versions.

composer.json
```diff
@@ -16,11 +16,11 @@
 
     "require": {
         "php":                   ">=5.3.1",
-        "symfony/css-selector":  "~2.1|~3.0"
+        "symfony/css-selector":  "^2.7|^3.0|^4.0"
     },
 
     "require-dev": {
-        "symfony/phpunit-bridge": "~2.7|~3.0"
+        "symfony/phpunit-bridge": "^3.3|^4.0"
```
aik099: You've increased `2.1` to `2.7` for `symfony/css-selector`. Why not doing something similar for 
 `symfony/phpunit-bridge` as well?

dunglas: The feature allowing to run both PHPUnit major versions has been added in 3.3, it doesn't work with earlier versions.

aik099: I guess Symfony 3.x works fine on PHP 5.3 it Travis CI is happy.

.travis.yml
```diff
@@ -2,28 +2,32 @@ language: php
...
 
 matrix:
   fast_finish: true
   include:
     - php: 5.3
+      dist: precise
+    - php: 7.1
       env: COMPOSER_FLAGS='--prefer-lowest --prefer-stable' SYMFONY_DEPRECATIONS_HELPER=weak
-    - php: 5.6
+    - php: 7.1
```
aik099: Is it needed to specify `-php 7.1` several times? You can combine multiple env vars into single line as well (above `COMPOSER_FLAGS` and `SYMFONY_DEPRECATIONS_HELPER` are combined into single line).

stof: Well, we want separate jobs for the lowest deps and the unstable deps (these cannot be combined together, as they are about different versions of deps).

and if you don't specify the PHP version for a job, Travis defaults to 5.5.

aik099: I see. So that was on purpose. Thanks for explaining.

aik099:
Now it seems PR is ready for mering. I can merge as well, but I need a confirmation, that there is all done what should be done.

composer.json
stof: the PHPUnit bridge is special. We kept PHP 5.3 support in it (allowing Symfony 2.7 to use the latest bridge version for its own testing, instead of having to backport nice changes all the time).

.travis.yml
stof: @aik099 Travis CI gives you a version of PHPUnit compatible with the PHP runtime. but when it gives you PHPUnit 4.8, it might give you a version older than 4.8.35, and so without the new namespaced class names for TestCase, making hard to have a testsuite compatible with PHPUnit 6 as well (which is used by default on Travis for PHP 7.2 for instance)

stof: using a phpunit version installed through composer could also be working. But the drawback is that it would install an old unmaintained version of phpunit when using `--prefer-lowest`. And this does not make sense (we want to test the lowest versions of our deps, but we can still use an uptodate testing tool for that, avoiding to be hit by old PHPUnit bugs)

.travis.yml
stof: please add PHP 7.2 in this list.

And I would remove HHVM, as they decided to not care about PHP parity anymore.

stof:
hmm, the Symfony PHPUnit Bridge installs PHPunit 6 on PHP 7.2 (as previous versions are triggering PHP deprecations on it). So it requires switching to namespaced PHPunit TestCase class.

dunglas:
@stof can I do that in another PR (and drop the last commit for now)?

stof:
@dunglas yes, sure

dunglas:
done

aik099:
Merging, thanks @dunglas .

sroze:
A release would be 🎆

aik099:
I thought, that this could not be as easy as just merging the PR.

@sroze , shouldn't you be using something like `"behat/mink":                "~1.7@dev",` in your `composer.json`. This way you'll be getting latest dev code for testing.

This is how all drivers are connected to Mink.

PR #751, opened 2017-12-04
Allow installation with Symfony 4

ciaranmcnulty:


ciaranmcnulty:
D'oh this is a huge duplication

PR #755, opened 2018-01-07
Add support for PHP 7.2

stof:
This also drops CI for HHVM due to https://hhvm.com/blog/2017/09/18/the-future-of-hhvm.html. Spending time on fixing HHVM issues is not worth it anymore (thus, Travis does not give us access to the latest HHVM)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,7 +2,7 @@ language: php
 
 sudo: false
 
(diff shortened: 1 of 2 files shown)


acrobat:
@stof Is it possible to tag a new release so we have a tag with sf4 and php 7.2 support? Thanks!

PR #760, opened 2018-03-27
Objects with __toString() can be asserted.

tamasd:
In a recent change, some regex matches changed to stripos() calls.
The problem is that stripos() converts the needle parameter to int
if it is not a string. While objects having __toString() can be
converted into string by most of the internal PHP functions, they
cannot be converted to int, resulting in an error.

I found this issue when running Drupal tests. Drupal passes objects to these assert functions (most notable `TranslatableMarkup`.


stof:
It converts it to int ? WTF. The function expects a string, not an int.

tests/WebAssertTest.php
```diff
@@ -255,7 +256,7 @@ public function testResponseHeaderContains()
                 )
             ));
 
-        $this->assertCorrectAssertion('responseHeaderContains', array('foo', 'ba'));
+        $this->assertCorrectAssertion('responseHeaderContains', array('foo', new Stringer('ba')));
```
stof: please create a *new* test instead of altering the existing one. We still want to test with an actual string, as that's the documented API.

tests/helpers/Stringer.php
```diff
@@ -0,0 +1,42 @@
+<?php
+
+namespace Behat\Mink\Tests\helpers;
```
stof: We always use singular names for our namespaces, and we start them with a capital letter

tests/helpers/Stringer.php
```diff
@@ -0,0 +1,42 @@
...
+
+/**
+ * Class Stringer
```
stof: this short description is totally useless (your long description should become the short one)

tests/helpers/Stringer.php
```diff
@@ -0,0 +1,42 @@
...
+ */
+class Stringer
+{
+
+  /**
+   * Internal storage.
+   *
+   * @var string
+   */
+  protected $content;
```
stof: should be private

tests/helpers/Stringer.php
```diff
@@ -0,0 +1,42 @@
...
+
+  /**
+   * Stringer constructor.
+   *
+   * @param $content
```
stof: invalid phpdoc (missing type)

tests/helpers/Stringer.php
```diff
@@ -0,0 +1,42 @@
...
+   */
+  public function __construct($content)
```
stof: please use 4 spaces for the indentation, according to our coding standards

aik099:
The `responseHeaderContains` method and others expect string as 2nd argument. Fact, that Drupal tests are passing object in there with expectation for it to be converted to a string by some magic surely violates proper function usage.

This was already discussed in #743, where `preg_match` was replaced by `stripos`.

tamasd:
@aik099 A little backstory on how I found this issue. I was developing tests for a Drupal module that we are developing internally. The test called traits from the Drupal core, and those traits asserted translated strings (which are objects). I looked around and it seems that the Drupal core does this in a lot of places.

Drupal core also depends on the master branch of mink for some reason. The easier fix would be to make Drupal depend on the last stable version, however that is a temporary fix, because a new mink version going to be released eventually.

I think adding a cast is much simplier than requiring bigger projects to go through bigger refactorings. And since many PHP functions convert object with `__toString()` to strings silently, it is not unreasonable to expect this behavior.

Also, PHP 7 does this conversion automatically in case when a function parameter is type hinted. See here: https://3v4l.org/vthrq

mxr576:
@aik099 Could you please take a look at this PR. I am not sure whether the failed test on Travis is actually a blocker of the merge or not.

aik099:
@mxr576 , checked. All issues have been resolved now.

aik099:
Sure. Merged now. Thanks @tamasd .

PR #766, opened 2018-10-18
Add hasOption method to NodeElement class.

kukko:
In my case that was a useful method, to check that a select element has an option or not. And I think this will be also useful for others.

### DIFF
--- a/src/Element/NodeElement.php
+++ b/src/Element/NodeElement.php
@@ -238,7 +238,7 @@ class NodeElement extends TraversableElement
     }
 
     /**
-     * Returns that the element has the find option or not.
+     * Returns that the option is in the element or not.
      * 
      * @param string $option
      * @return Boolean
--- a/src/Element/NodeElement.php
+++ b/src/Element/NodeElement.php
@@ -238,6 +238,24 @@ class NodeElement extends TraversableElement
     }
 
     /**
+     * Returns that the element has the find option or not.
+     * 
+     * @param string $option
+     * @return Boolean
+     */
+    public function hasOption($option){
+        $previousValue=$this->getValue();
+        try{
+            $this->selectOption($value);
+            $this->setValue($previousValue);
+            return true;
+        }
+        catch (ElementNotFoundException $e){
+            return false;
+        }
+    }
+
+    /**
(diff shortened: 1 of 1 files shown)


stof:
This will trigger 2 changes of the data of the form, to check whether it has an option or no. that's quite bad as implementation of the check.

kukko:
Yeah you are right with this. I will rewrite it to don't change the value of the select.

src/Element/NodeElement.php
```diff
@@ -237,6 +237,24 @@ public function selectOption($option, $multiple = false)
         $this->getDriver()->selectOption($this->getXpath(), $opt->getValue(), $multiple);
     }
 
+    /**
```
aik099: There are multiple CS issues in the code, like missing spaces.

kukko: I fixed the missing spaces before and after the equality signs and the logical expressions.

src/Element/NodeElement.php
```diff
@@ -237,6 +237,20 @@ public function selectOption($option, $multiple = false)
...
+     * Returns that the option is in the element or not.
+     * 
+     * @param string $option
+     * @return Boolean
+     */
+    public function hasOption($option){
+        if ('select' !== $this->getTagName()) {
+            return;
```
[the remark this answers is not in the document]
kukko: I writed this because if the element is not a select tag, It is not logical to return a boolean. That would be nice if I throw an exception instead of return null? Or editing the documentation of method to `@return Boolean|null`?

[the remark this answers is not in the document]
aik099: IMO it's better to throw an exception

src/Element/NodeElement.php
```diff
@@ -237,6 +237,23 @@ public function selectOption($option, $multiple = false)
...
+     * @param string $option
+     * @return Boolean|null
+     */
+    public function hasOption($option)
+    {
+        if ('select' !== $this->getTagName()) {
+            return;
+        }
+
+        $optionElement = $this->find('named', array('option', $option));
```
aik099: You can safely inline this variable, because it's used only once.

PR #767, opened 2018-10-18
Add setAutocompleteValue method

JordiGiros:
Added method in order to solve setting a value on autocomplete input elements as commented in [286](https://github.com/minkphp/MinkSelenium2Driver/pull/286), [292](https://github.com/minkphp/MinkSelenium2Driver/issues/292) and  [301](https://github.com/minkphp/MinkSelenium2Driver/issues/301)

### DIFF
--- a/src/Driver/DriverInterface.php
+++ b/src/Driver/DriverInterface.php
@@ -366,6 +366,19 @@ interface DriverInterface
     public function setValue($xpath, $value);
 
     /**
+     * Sets autocomplete element's value by it's XPath query.
+     *
+     * @param string            $xpath
+     * @param string|bool|array $value
+     *
+     * @throws UnsupportedDriverActionException When operation not supported by the driver
+     * @throws DriverException                  When the operation cannot be done
+     *
+     * @see \Behat\Mink\Element\NodeElement::setAutocompleteValue
+     */
+    public function setAutocompleteValue($xpath, $value);
+
+    /**
      * Checks checkbox by it's XPath query.
      *
      * @param string $xpath
--- a/src/Driver/CoreDriver.php
+++ b/src/Driver/CoreDriver.php
@@ -180,6 +180,14 @@ abstract class CoreDriver implements DriverInterface
     /**
      * {@inheritdoc}
      */
+    public function setAutocompleteValue($xpath, $value)
+    {
+        throw new UnsupportedDriverActionException('Setting the field value is not supported by %s', $this);
+    }
+
+    /**
+     * {@inheritdoc}
+     */
     public function check($xpath)
     {
         throw new UnsupportedDriverActionException('Checking a checkbox is not supported by %s', $this);


stof:
I don't think the method name is right. This method should be about typing in a field, not about setting a value IMO. It is not only about autocomplete

stof:
Note that if we make it a method typing something in the field, we should probably not force emptying first (as that removes control from the user about what gets actually typed in the field)

stof:
This would make it map to https://www.w3.org/TR/webdriver1/#element-send-keys

JordiGiros:
> Note that if we make it a method typing something in the field, we should probably not force emptying first (as that removes control from the user about what gets actually typed in the field)

Emptying first as in setValue method? Why not? The behaviour should be nearly the same but without triggering any "change event" and staying the focus in the element. Does this apply here or is a problem we should discus in MinkSelenium2 project?

JordiGiros:
> I don't think the method name is right. This method should be about typing in a field, not about setting a value IMO. It is not only about autocomplete

Ok, I agree. We can name it "sendKeys()" following W3 naming.

stof:
@JordiGiros if you test things for a complex JS widget, you might want to see what happens when writing `fr` in the field, and then continuing writing to reach `fran`. If we force to *set* the value, emptying the field first (which is what https://github.com/minkphp/MinkSelenium2Driver/pull/302 is doing currently), you cannot test that (as your second call would erase everything and then write `fran`, which is a different usage pattern).
As the intended usage of this method is to be able to test such JS-based UIs, we should make  it as usable as possible for that.

stof:
We also need a method on NodeElement

stof:
And we need a PR on the driver-testsuite repository with functional tests using this new API, to cover the expected behavior of drivers (otherwise, there is no interoperability between drivers)

JordiGiros:
Yes, I see. So not deleting previous string and using the keyPress() in loop would be a better solution. I'll change that.

paulbriton:
Any update on this?

aik099:
> https://github.com/minkphp/Mink/pull/767#issuecomment-431032401

@JordiGiros , I've specified a place in https://github.com/minkphp/MinkSelenium2Driver/pull/302 that needs to be changed to impelemet this.

src/Driver/DriverInterface.php
```diff
@@ -365,6 +365,19 @@ public function getValue($xpath);
...
+    /**
+     * Sends keys to an element by it's XPath query.
+     *
+     * @param string            $xpath
+     * @param string|bool|array $value
+     *
+     * @throws UnsupportedDriverActionException When operation not supported by the driver
+     * @throws DriverException                  When the operation cannot be done
+     *
+     * @see \Behat\Mink\Element\NodeElement::setAutocompleteValue
```
aik099, line 377: Are you sure, that mentioned method exists?

aik099 (changes requested):

PR #768, opened 2018-10-26
Move hasContent method from DocumentElement class to TraversableElement.

kukko:
It's useful to find content in DOM elements, not just in the whole document.


aik099:
What about tests. I guess now we also need to check, that `hasContent` method of all descendants from `TraversableElement` class works, the `NodeElement` maybe.

PR #774, opened 2019-05-14
Rename boolean type in lowercase

loic425:
vimeo/psalm raise errors from Behat/Mink boolean return type.
As I can see in a [psalm issue](https://github.com/vimeo/psalm/issues/294#issuecomment-344560596), Boolean type must be lowercase

### DIFF
--- a/src/Driver/DriverInterface.php
+++ b/src/Driver/DriverInterface.php
@@ -53,7 +53,7 @@ interface DriverInterface
     /**
      * Checks whether driver is started.
      *
-     * @return Boolean
+     * @return boolean
      */
     public function isStarted();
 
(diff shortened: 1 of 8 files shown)


aik099:
Some questions:

1. no such issues with other data types?
2. have you checked driver repositories for the same issue?
3. how you've managed to see that issue, because Mink is a dev dependency and usually isn't analyzed by anything?

loic425:
> Some questions:
> 
> 1. no such issues with other data types?
> 2. have you checked driver repositories for the same issue?
> 3. how you've managed to see that issue, because Mink is a dev dependency and usually isn't analyzed by anything?

1. I don't think so
2. Nope
3. I managed to see that issue during that [Monofony pull request ](https://github.com/Monofony/SymfonyStarter/pull/180). Cause Behat classes are on src directory (same directory on Sylius).

stof:
it would be great if you could provide the same work for drivers

loic425:
> it would be great if you could provide the same work for drivers

I think it's done. I check all the active drivers' repositories. I don't have check deprecated ones.

loic425:
@stof This one is merged, but the pull-requests on active drivers are not. Is there a reason ?

Issue #696, opened 2016-03-08
Assert non-empty field value


lauft:
Assume an text input field in a form:

```
<input type="text" value="" name="foo" id="foo">
```

When I try to assert that the field is not empty using `WebAssert::fieldValueNotEquals('foo', '')` it fails because the function internally converts the parameter `$value` to the regex `/^/ui` which matches every string.

A workaround is to call `WebAssert::fieldValueNotEquals('foo', '\n')` but that is somehow counterintuitive...

Of course, I could check for the precise value of the input field, but there are cases where I do not know or care about this value...

aik099:
Sounds like a bug. Could you please send PR, that updates both `fieldValueEquals` and `fieldValueNotEquals` to do exact match (the `===`) instead of `preg_match` when empty value is given for comparison.

lauft:
The issue with the empty field value seems to be fixed in the current version.
However, during my investigation I think I discovered some further issues with the WebAssert functions fieldValueEquals and fieldValueNotEquals.

The two main things were
- the used preg_match compared case insensitive. I think this is wrong, as text input 'foo' should not be considered equal to 'FOO', 'Foo', etc.
- the function Behat\Mink\Element\NodeElement::getValue() returns string|bool|array according to PHPDoc. The test functions for fieldValueEquals and fieldValueNotEquals used integers to check against.

I created and sent a pull request (https://github.com/Behat/Mink/pull/2) for these issues to review.

Concerning the first issue in the above list, I think I might add functions fieldValueMatches/fieldValueNotMatches which use regular expression with another pull request.

aik099:
The PR should be send to https://github.com/minkphp/Mink instead of https://github.com/Behat/Mink repo.

aik099:
The https://github.com/Behat/Mink is fork from https://github.com/minkphp/Mink made to allow downloading older Mink releases prior it's move from Behat into minkphp organization.

lauft:
I am going to rework the pull-request to serve both requirements of Behat/Mink and minkphp/Mink (i.e. splitting it in two). The one for minkphp needs rebasing anyway as you pointed out.

aik099:
To avoid confusion:

1. close PR to Behat/Mink repository (because PR should go into minkphp/Mink repository)
2. incorporate changes described on Behat/Mink PR into PR that goes into minkphp/Mink

aik099:
@lauft , any update on PR status?

lauft:
[...]

PR #775, opened 2019-05-24
Add fieldValueMatches / fieldValueNotMatches

das-peter:
Would be hand to allow pattern assertions on field values like on the page text.
This was mentioned here #696 but: "PR got lost in the sands of time" :)


das-peter:
Added tests, based on testPageTextMatches / testPageTextNotMatches / testFieldValueEquals / testFieldValueNotEquals

aik099:
Looks good to me.

PR #786, opened 2019-09-04
Changelog for 1.7.2

greg-1-anderson:
Preliminary.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,3 +1,33 @@
+1.7.2 / 2019-09-XX
+==================
+
+Bug fixes:
+
+* DriverInterface#getValue() can return null
+* Modify response[Contains|NotContains] to use strpos vs preg_match
+* Fix compatibility with PHP 7.2
+* Allow Symfony 4.
+* Optimize the splitting of XPath unions
+* Allow pipes (|) in attributes by implementing better XPath parsing
+* Auto-start session only upon 1st "->visit(...)" method call
+
+Testsuite:
+
+* Don't install PHPUnit 8.0
+* Fix PHP 5.4 and PHP 5.5 builds
+* Objects with __toString() can be asserted.
+* Add testing on PHP 7.2
+* Add compatibility with PHPUnit 6
+* Remove the driver testsuite
+
+Misc:
+
+* Rename boolean type in lowercase
+* Suggest chrome-mink-driver as fast JS-enabled alternative to selenium
+* Use ! operator instead of === false
+* Improved variable names and simplified quote logic
+* Changed method visibility to private and variable name to camle case
+
 1.7.1 / 2016-03-05
 ==================
 


CHANGES.md
```diff
@@ -1,3 +1,33 @@
...
+* Auto-start session only upon 1st "->visit(...)" method call
+
+Testsuite:
+
+* Don't install PHPUnit 8.0
+* Fix PHP 5.4 and PHP 5.5 builds
+* Objects with __toString() can be asserted.
+* Add testing on PHP 7.2
+* Add compatibility with PHPUnit 6
+* Remove the driver testsuite
```
oleg-andreyev: I would rephrase it: Extracted driver test suite into separate repository.

CHANGES.md
```diff
@@ -1,3 +1,33 @@
+1.7.2 / 2019-09-XX
+==================
+
+Bug fixes:
+
+* DriverInterface#getValue() can return null
+* Modify response[Contains|NotContains] to use strpos vs preg_match
+* Fix compatibility with PHP 7.2
+* Allow Symfony 4.
+* Optimize the splitting of XPath unions
```
aik099: Please remove this one.

---

Historically there were 2 PRs:

* in 1st one support for unions was added
* in 2nd one that support was improved

For end user these looks as a single change since it happened inside same version.

CHANGES.md
```diff
@@ -1,3 +1,33 @@
...
+* Fix PHP 5.4 and PHP 5.5 builds
+* Objects with __toString() can be asserted.
+* Add testing on PHP 7.2
+* Add compatibility with PHPUnit 6
+* Extracted driver test suite into separate repository.
+
+Misc:
+
+* Rename boolean type in lowercase
+* Suggest chrome-mink-driver as fast JS-enabled alternative to selenium
```
aik099: Please remove this line.

---

This is an internal change, that shouldn't be affecting developer workflow.

CHANGES.md
```diff
@@ -1,3 +1,33 @@
...
+* Allow Symfony 4.
+* Optimize the splitting of XPath unions
+* Allow pipes (|) in attributes by implementing better XPath parsing
+* Auto-start session only upon 1st "->visit(...)" method call
+
+Testsuite:
+
+* Don't install PHPUnit 8.0
+* Fix PHP 5.4 and PHP 5.5 builds
+* Objects with __toString() can be asserted.
```
aik099: Please move into `New Features:` sub-section and change text into:

```
* The `WebAssert` class methods now accept objects with `__toString` method, where only string were allowed before
```

---

The `WebAssert` class is part of public API and not the test suite.

CHANGES.md
```diff
@@ -1,3 +1,33 @@
...
+* Add testing on PHP 7.2
+* Add compatibility with PHPUnit 6
+* Extracted driver test suite into separate repository.
+
+Misc:
+
+* Rename boolean type in lowercase
```
aik099: Please remove this line.

---

This is an internal change, that shouldn't be affecting developer workflow.

CHANGES.md
```diff
@@ -1,3 +1,33 @@
+1.7.2 / 2019-09-XX
+==================
+
+Bug fixes:
+
+* DriverInterface#getValue() can return null
```
aik099: Please remove this line.

---

The method returned NULL before, but that wasn't mentioned in the DocBlock. DocBlock change isn't something, that goes into change log ;)

CHANGES.md
```diff
@@ -1,3 +1,33 @@
...
+* Modify response[Contains|NotContains] to use strpos vs preg_match
+* Fix compatibility with PHP 7.2
+* Allow Symfony 4.
+* Optimize the splitting of XPath unions
+* Allow pipes (|) in attributes by implementing better XPath parsing
+* Auto-start session only upon 1st "->visit(...)" method call
+
+Testsuite:
+
+* Don't install PHPUnit 8.0
```
aik099: Please remove this line.

---

What PHPUnit versions are used to test library itself doesn't affect end users of the library.

CHANGES.md
```diff
@@ -1,3 +1,33 @@
...
+* Fix PHP 5.4 and PHP 5.5 builds
```
aik099: Please change into:

```
* Fixed PHP 5.4 and PHP 5.5 builds
```

CHANGES.md
```diff
@@ -1,3 +1,33 @@
...
+* Objects with __toString() can be asserted.
+* Add testing on PHP 7.2
+* Add compatibility with PHPUnit 6
```
aik099: Please remove this line.

---

What PHPUnit versions are used to test library itself doesn't affect end users of the library.

CHANGES.md
```diff
@@ -1,3 +1,33 @@
...
+* Optimize the splitting of XPath unions
+* Allow pipes (|) in attributes by implementing better XPath parsing
+* Auto-start session only upon 1st "->visit(...)" method call
+
+Testsuite:
+
+* Don't install PHPUnit 8.0
+* Fix PHP 5.4 and PHP 5.5 builds
+* Objects with __toString() can be asserted.
+* Add testing on PHP 7.2
```
aik099: Please change into:

```
* Added build for PHP 7.2
```

CHANGES.md
```diff
@@ -1,3 +1,33 @@
...
+* Add compatibility with PHPUnit 6
+* Extracted driver test suite into separate repository.
+
+Misc:
+
+* Rename boolean type in lowercase
+* Suggest chrome-mink-driver as fast JS-enabled alternative to selenium
+* Use ! operator instead of === false
```
aik099: Please remove this line.

---

This is an internal change, that shouldn't be affecting developer workflow.

CHANGES.md
```diff
@@ -1,3 +1,33 @@
...
+* Improved variable names and simplified quote logic
```
aik099: Please remove this line.

---

This is an internal change, that shouldn't be affecting developer workflow.

CHANGES.md
```diff
@@ -1,3 +1,33 @@
...
+* Changed method visibility to private and variable name to camle case
```
aik099: Please remove this line.

---

This is an internal change, that shouldn't be affecting developer workflow.

CHANGES.md
stof: this one should be listed IMO, as it precisely help developers running Psalm on a project using Mink (that was the reason for the change)

CHANGES.md
stof: and having a `new features` section also means this cannot be a patch release.

CHANGES.md
```diff
@@ -1,3 +1,33 @@
...
+
+Bug fixes:
+
+* DriverInterface#getValue() can return null
+* Modify response[Contains|NotContains] to use strpos vs preg_match
+* Fix compatibility with PHP 7.2
+* Allow Symfony 4.
+* Optimize the splitting of XPath unions
+* Allow pipes (|) in attributes by implementing better XPath parsing
+* Auto-start session only upon 1st "->visit(...)" method call
```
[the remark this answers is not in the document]
stof: I would actually put this in a `New features` section.

CHANGES.md
```diff
@@ -1,3 +1,33 @@
+1.7.2 / 2019-09-XX
+==================
+
+Bug fixes:
+
+* DriverInterface#getValue() can return null
+* Modify response[Contains|NotContains] to use strpos vs preg_match
```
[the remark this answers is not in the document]
stof: should be removed IMO. This is a change of an internal implementation, without any user-facing impact.

CHANGES.md
```diff
@@ -1,3 +1,33 @@
...
+* Auto-start session only upon 1st "->visit(...)" method call
+
+Testsuite:
+
+* Don't install PHPUnit 8.0
+* Fix PHP 5.4 and PHP 5.5 builds
+* Objects with __toString() can be asserted.
+* Add testing on PHP 7.2
+* Add compatibility with PHPUnit 6
+* Extracted driver test suite into separate repository.
```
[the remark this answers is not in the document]
greg-1-anderson: Removed all periods.

CHANGES.md
aik099: Isn't that the case, where `vendor` folder was analyzed by mistake 😄 . Let's leave it then.

CHANGES.md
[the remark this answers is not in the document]
aik099: I wanted to keep it because I remember discussions, where person was using API in incorrect way (knowing that `preg_match` is used internally and specified regular expressions in the input) started to complain about change to `stripos`, because his tests started to fail for no reason.

If that is no longer the case, then we can safely remove this line.

CHANGES.md
aik099: Ups. Then either whole release needs to wait for #772 or we should release 1.8.0 now and #772 will go into 1.9.0 only.

greg-1-anderson:
Updated per review comments.

One request, if I may. Similar to the situation in https://github.com/minkphp/MinkSelenium2Driver/issues/313, Drupal is also depending on the 1.7.x-dev branch alias for this repository. Could you please create an unmaintained `1.7` branch before changing the branch alias to 1.8.x-dev?

Sorry for the inconvenience.

CHANGES.md
stof: to analyze your own code *using Mink*, psalm needs to know the types of Mink APIs. And so it will look at the phpdoc (and it seems it considers `Boolean` as a class name)

aik099:
@greg-1-anderson , not all items were addressed. I've resolved conversations for items that were fixed. Others needs to be addressed.

greg-1-anderson:
@aik099 : I addressed most of the items that remained unresolved. The exceptions:

```
Objects with __toString() can be asserted
```
> and having a new features section also means this cannot be a patch release.

I already moved this line up to "features" and bumped the expected version to 1.8.0, but this comment remains unresolved. Was there something else you wanted to happen here?

```
Modify response[Contains|NotContains] to use strpos vs preg_match
```
> should be removed IMO. This is a change of an internal implementation, without any user-facing impact

Already removed, but this not marked resolved. A mistake, perhaps?

Please let me know if you want any further changes with these items or anything else.

aik099:
> Objects with __toString() can be asserted

Resolved.

> Modify response[Contains|NotContains] to use strpos vs preg_match

Could you bring it back (the fixed version). I consider this important change.

Now we're waiting for #787, where @stof suggests to make a minor release of every mink repo with current features used for Drupal and then do a feature release, where available.

stof:
> Now we're waiting for #787, where @stof suggests to make a minor release of every mink repo with current features used for Drupal and then do a feature release, where available.

I don't understand what you mean. minor versions are precisely the versions receiving features in semver.

stof:
> > Modify response[Contains|NotContains] to use strpos vs preg_match
> 
> Could you bring it back (the fixed version). I consider this important change.

Why ? that's an internal performance optimization to avoid using a regex when not necessary. The behavior is exactly the same.

aik099:
> https://github.com/minkphp/Mink/pull/786#issuecomment-528796100

@stof , my bad. I thought, that features goes into `major` and bugs go into `minor`. You're correct. Now I see.

> https://github.com/minkphp/Mink/pull/786#issuecomment-528796352

@stof , if people use these methods as expected by us, then yes, nothing is changed. However they could have passed-in regex fragments before and that worked as well. That won't work anymore now.

stof:
>  if people use these methods as expected by us, then yes, nothing is changed. However they could have passed-in regex fragments before and that worked as well. That won't work anymore now.

this was not possible before either. We were properly escaping their input with `preg_quote`.

greg-1-anderson:
Based on the above, it sounds to me as if the preg_match removal was not a significant change. I'm ready to bring that comment back if you decide you want it, though.

aik099:
> https://github.com/minkphp/Mink/pull/786#issuecomment-528851142

@stof , fine. Then it's good, that @greg-1-anderson removed it.

aik099:
@greg-1-anderson , since all what I've requested was done I'm approving the PR.

scaytrase:
Any news here?

neclimdul:
Is this still happening? Would be really great to have a new release.

aik099:
I guess we're waiting for @stof review and then we can merge it. Not sure if anything else is blocking the release though.

alexpott:
It would be really really great to have PHP 7.4 compatibility in this release. The smallest fix for that is #792 but there's also #794

aik099:
> It would be really really great to have PHP 7.4 compatibility in this release. The smallest fix for that is #792 but there's also #794

@alexpott , I've approved #792 (@stof can merge it unless he has) and commented upon #794.

stof:
I ended up doing my own update of the changelog to make the release. Some of the bug fixes listed in this changelog are actually fixes compared to the previous PR merged, not fixes compared to the previous release, and so there is no point highlighting them in the changelog (the entry would read `we avoided regressions when refactoring` if comparing to the previous release).

PR #788, opened 2019-09-26
add php versions

andreybolonin:


stof:
we now have PHP 7.3 and 7.4 on Travis through other PRs

PR #789, opened 2019-10-11
Allow selectors and replacements to be registered late

andrewnicols:
This allows projects using Mink to create their Named and Partial
selectors, and then to add additional replacements and selectors to be
registered for subsequent use.

This is particularly useful in projects with a distributed plugin
architecture which make use of replacements and selectors
without abusing Reflection to do so.

This is something that we currently do in Moodle for Behat for exactly this reason:
Our hack with Reflection to access the properties to allow proper translation: https://github.com/andrewnicols/moodle/blob/MDL-66559-master/lib/behat/classes/named_selector.php#L48
Registration of plugin replacements and selectors: https://github.com/andrewnicols/moodle/blob/MDL-66559-master/lib/tests/behat/behat_hooks.php#L743

### DIFF
--- a/tests/Selector/NamedSelectorTest.php
+++ b/tests/Selector/NamedSelectorTest.php
@@ -59,6 +59,37 @@ abstract class NamedSelectorTest extends TestCase
     }
 
     /**
+     * @dataProvider getLateRegisteredReplacements
+     */
+    public function testLateRegisteredReplacements($fixtureFile, $replacements, $selectors, $selector, $locator, $expectedExactCount, $expectedPartialCount = null)
+    {
+        $expectedCount = $this->allowPartialMatch() && null !== $expectedPartialCount
+            ? $expectedPartialCount
+            : $expectedExactCount;
+
+        // Don't use "loadHTMLFile" due HHVM 3.3.0 issue.
+        $dom = new \DOMDocument('1.0', 'UTF-8');
+        $dom->loadHTML(file_get_contents(__DIR__.'/fixtures/'.$fixtureFile));
+
+        $namedSelector = $this->getSelector();
+
+        foreach ($replacements as $from => $to) {
+            $namedSelector->registerReplacement($from, $to);
+        }
+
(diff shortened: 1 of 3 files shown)


andrewnicols:
Is there anything that we can do to see this land?

It's extremely helpful and powerful to be able to add additional locators and replacements to the various Selector classes, and being able to make use of Mink replacements in our own named selectors in this way is a huge benefit.

tests/Selector/NamedSelectorTest.php
```diff
@@ -162,6 +193,56 @@ public function getSelectorTests()
         );
     }
 
+    public function getLateRegisteredReplacements()
+    {
+        // Fixture file,  selectors,  replacements, selector name,  locator,  expected number of exact matched elements, expected number of partial matched elements if different
+        return array(
```
aik099, line 199: With all these replacements in place, I'm not clearly seeing from the test code what each fixture should find (what the final selector for search should look like). 

Please add some comments about each fixture with selectors .

tests/Selector/NamedSelectorTest.php
```diff
@@ -162,6 +193,56 @@ public function getSelectorTests()
...
+            'lateSelectorUsingLateReplacement' => array(
```
aik099, line 200: With all these replacements in place, I'm not clearly seeing from the test code what each fixture should find (what the final selector for search should look like). 

Please add some comments about each fixture with the CSS selectors, that are supposed to be found.

aik099 (changes requested):

andrewnicols:
I've updated the unit test to more clearly document the provided data an expectations.

I hope that this clarifies things. Please let me know if you can think of a better solution to this.

Thanks

andrewnicols:
Note: Travis tests are failing because xdebug is not enabled by default for the coverage reports.

aik099 (approved):

aik099:
Merging, thank you @andrewnicols for a fast response.

Issue #787, opened 2019-09-05
[RFC] New PHP min version for Mink and drivers


stof:
There is some discussions happening on various driver repos about dropping PH 5.3 support. This issue is about discussing a consolidated plan for the whole set of Mink repos, to be consistent.

Arguments to drop support:

- all PHP versions < 7.1 are unmaintained (and PHP 7.1 reaches that point at the end of the year)
- Travis supports only PHP 5.6+ on their xenial infrastructure, forcing to use trusty for 5.4 and 5.5 and precise for 5.4. And they announced that the precise infrastructure will be decommissioned in the future, forbidding us to run tests on 5.3 at that point (and so to officially support it)
- PHPUnit is dropping support for unmaintained PHP versions regularly, which forces us to support a wide range of PHPUnit versions to run our testsuite, which becomes more painful over time. For now, we need these versions:
    - 4.8 for PHP 5.3 to 5.5
    - 5.7 for PHP 5.6
    - 6.5 for PHP 7.0
    - 7.4 for PHP 7.1+
    - we are currently not yet compatible with the new PHPUnit 8, but it supports PHP 7.2+
- PHPUnit 8 is adding `void` return type in many places. Making our testsuite compatible with PHPUnit 8 will be a pain if we need to keep support for PHP < 7.1
- Symfony 5 will add return types on a some interfaces, which will force us to do the same in our testsuite implementation (for BrowserKitDriver), which will require dropping support for PHP 5.x (or do crazy stuff to keep 2 implementations). One of the methods affecting us will use the `object` type, which requires PHP 7.2
- our new driver based on facebook/webdriver will require at least PHP 5.6+ (due to the min requirement of the library and they plan to make it PHP 7.1+ in the next version). This is not a strict requirement to match the other parts of Mink (it would have this requirement even without this min version bump on Mink), but it is nice to take that into account in the thinking.

As we start dropping support for old versions, I suggest dropping more than just 5.3, to benefit from a cleanup (as we don't do it often). I suggest to at least drop 5.3 to 5.5, which will free us from old Travis infrastructure and from PHPUnit 4.8. Then, multiple suggestions are possible:

- 7.2+, to make it easy to add support for Symfony 5 (but this is mostly things happening in the testsuite, so we could deal with it otherwise)
- 7.1+, to make it easy to migrate to PHPUnit 8 (while keeping support for PHPUnit 7.4 due to our needs), which also aligns with the min requirement of Symfony LTS, and with the current maintained PHP versions
[...]

PR #791, opened 2019-10-30
Include symfony 5. Corections for phpunit 9

breizh81:
#hackdayparis

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -5,6 +5,7 @@
     "homepage":     "http://mink.behat.org/",
     "type":         "library",
     "license":      "MIT",
+    "minimum-stability": "dev",
 
     "authors": [
         {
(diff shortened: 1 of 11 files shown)


aik099:
This is a BC break. Please see #787.

stof:
This PR is not about supporting Symfony 5 (which is already supported without these changes, but only changing the `css-selector` constraint), but about supporting PHPUnit 9 in our testsuite (due to `symfony/phpunit-bridge` selecting it by default).
I refactored our CI setup to have better control over the PHPUnit version (so that we decide when we update it, not Symfony deciding it for us). Using PHPUnit 9 will be done only after dropping support for older PHP versions (which is my next task on Mink now that releases are done), which will be much simpler than what should be done here to have it green. So I'm closing this PR.

PR #792, opened 2019-11-20
Php73 74

stronk7:
This adds travis testing of php73 and php74 and fixes the only problem we have detected with php74 so far.

Before: https://travis-ci.org/stronk7/Mink/builds/614672369
After: https://travis-ci.org/stronk7/Mink/builds/614678756

For your consideration, ciao :-)

PR #794, opened 2020-02-19
Add compatibility to symfony 5 and php 7.4

alexander-schranz:
related to #791, #792, #788

### DIFF
--- a/src/Selector/Xpath/Escaper.php
+++ b/src/Selector/Xpath/Escaper.php
@@ -47,6 +47,6 @@ class Escaper
             }
         }
 
-        return sprintf('concat(%s)', implode($parts, ','));
+        return sprintf('concat(%s)', implode(',', $parts));
     }
 }
--- a/.gitignore
+++ b/.gitignore
@@ -1,5 +1,6 @@
 *.tgz
 *.phar
 phpunit.xml
+.phpunit.result.cache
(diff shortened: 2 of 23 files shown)


sadikoff:
Any updates on it?

src/Selector/Xpath/Escaper.php
```diff
@@ -47,6 +47,6 @@ public function escapeLiteral($s)
             }
         }
 
-        return sprintf('concat(%s)', implode($parts, ','));
+        return sprintf('concat(%s)', implode(',', $parts));
```
aik099: This is already done as part of #792.

aik099:
@stof , changes made here is a variation of multi-PHPUnit version support, that was done prior to any decision reached in #787.

stof:
This PR is not about supporting PHP 7.4 or Symfony 5 (which are both already supported without these changes), but about supporting PHPUnit 9 (due to `symfony/phpunit-bridge` selecting it by default).
I refactored our CI setup to have better control over the PHPUnit version (so that we decide when we update it, not Symfony deciding it for us). Using PHPUnit 9 will be done only after dropping support for older PHP versions (which is my next task on Mink now that releases are done), which will be much simpler than what is done here. So I'm closing this PR.

PR #801, opened 2020-08-04
Improve responsivity (reduce fixed delay from 100 ms to 10 ms)

mvorisek:


aik099:
Why this is needed? Could you please explain the use case.

The current delay of 100000μs (100ms or 0.1sec), which is already quite small.

mvorisek:
> 
> 
> Why this is needed? Could you please explain the use case.
> 
> The current delay of 100000μs (100ms or 0.1sec), which is already quite small.

This can improve total test time by about 30% - the typical page load/ajax response time of fast webserver/app is in 30, 40 ms range and waiting for whole 100 ms after every test/wait is very, very inefficient. The same applies if page waiting condition is satisfied like in 110 ms - currently, we wait whole 200 ms (+45 %) instead of 110, 120 ms (+5%).

As this improves real world performance/responsivity with not too much overhead I think 10 ms delay is much better.

mvorisek:
> 
> 
> Why this is needed? Could you please explain the use case.
> 
> The current delay of 100000μs (100ms or 0.1sec), which is already quite small.

This can improve total test time by about 30% - the typical page load/ajax response time of fast webserver/app is in 30, 40 ms range and waiting for whole 100 ms after every test/wait is very, very inefficient. The same applies if page waiting condition is satisfied like in 110 ms - currently, we wait whole 200 ms (+45 %) instead of 110, 120 ms (+5%).

As this improves real world performance/responsivity with not too much overhead I think 10 ms delay is much better.

mvorisek:
@aik099 can this PR be merged?

aik099:
After @stof review, yes.

mvorisek:
> After @stof review, yes.

he reviewed it

can we merge it now and also do a release? When this function is called a lot, it has large speed benefit.

also https://github.com/minkphp/MinkSelenium2Driver/pull/322

aik099:
Merging, thanks @mvorisek .

mvorisek:
@aik099 @stof would it be possible to make a build version release because of this improvement?

mvorisek:
here are measured results with this patch: https://github.com/atk4/ui/pull/1501

PR #807, opened 2021-07-14
Don't put NULL into strtolower() and preg_match()

alexpott:
https://3v4l.org/KSvQ1/rfc#vgit.master

### DIFF
--- a/src/WebAssert.php
+++ b/src/WebAssert.php
@@ -219,7 +219,7 @@ class WebAssert
      */
     public function responseHeaderMatches($name, $regex)
     {
-        $actual = $this->session->getResponseHeader($name);
+        $actual = (string) $this->session->getResponseHeader($name);
         $message = sprintf('The pattern "%s" was not found anywhere in the "%s" response header.', $regex, $name);
 
         $this->assert((bool) preg_match($regex, $actual), $message);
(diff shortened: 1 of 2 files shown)


aik099:
@alexpott in which PHP version are you getting the deprecation notice shown on https://3v4l.org/KSvQ1/rfc#vgit.master ?

aik099:
Related to https://github.com/minkphp/MinkSelenium2Driver/pull/329 .

alexpott:
PHP 8.1

src/WebAssert.php
```diff
@@ -219,7 +219,7 @@ public function responseHeaderNotContains($name, $value)
      */
     public function responseHeaderMatches($name, $regex)
     {
-        $actual = $this->session->getResponseHeader($name);
+        $actual = (string) $this->session->getResponseHeader($name);
```
aik099, line 222: Why this (and similar type-casted places) is needed?

aik099 (changes requested):

src/WebAssert.php
alexpott: https://3v4l.org/hOfGr/rfc#vgit.master

https://php.watch/versions/8.1/internal-func-non-nullable-null-deprecation

aik099:
Merging, thank you @alexpott .

PR #808, opened 2021-07-16
Fix some more PHP 8.1 deprecations

alexpott:
Found a couple more places that a deprecation notice is triggered.


aik099:
Merging. Thank you @alexpott .

PR #809, opened 2021-08-31
Allow passing and setting basic auth on session.

### LINKED ISSUE #810: Feature: set basic auth after session start and reset
Testing sites with basic auth set became an issue for us after some extensions updates. 
In particular, an extension was calling the session reset which effectively removes the pre-set basic auth credentials (header). 

It would be nice if mink session would support setting basic auth credentials and use them if they are set after session start & reset. 

I've created a PR that fixes the issue for us. Please take a look and let us know if it would make sense to support this.
PR: https://github.com/minkphp/Mink/pull/809

petarbasic:
Resolves #810

### DIFF
--- a/src/Session.php
+++ b/src/Session.php
@@ -24,6 +24,8 @@ class Session
     private $driver;
     private $page;
     private $selectorsHandler;
+    private $basicAuthUser;
+    private $basicAuthPassword;
 
     /**
      * Initializes session.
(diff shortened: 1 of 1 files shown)


fago:
This came up at https://github.com/jhedstrom/drupalextension/issues/597 but would make sense to fix in mink actually.

aik099:
Closing, see associated discussion for a reason.

PR #812, opened 2021-10-03
Move to github actions

alexpott:


alexpott:
You see the new github actions working over here: https://github.com/alexpott/Mink/actions/runs/1300769865

aik099:
@alexpott , what problems have you encountered, that forced you to drop PHP 5.3, PHP 5.4, PHP 5.5 support?

The Setup PHP GitHub Action seems to support them: https://github.com/marketplace/actions/setup-php-action .

alexpott:
@aik099 we run out of simple-phpunit polyfil support. I think maybe we can re-enable 5.5 testing. See the logs for the 5.3 test here... https://github.com/alexpott/Mink/actions/runs/1304132764 - and yep 5.5 works...

That said I can work around this and re-enable it.

alexpott:
@aik099 a sea of green from PHP 5.3 to PHP 8.1  🍰 https://github.com/alexpott/Mink/actions/runs/1304192898

alexpott:
PR now uses Yoast's polyfills for BC and has testing on PHP 5.4 to PHP 8.1 using composer to control PHPUnit version... @stof how does this look?

.github/workflows/tests.yml
```diff
@@ -0,0 +1,64 @@
...
+
+      - name: Run tests
+        run: |
+          vendor/bin/phpunit -v --coverage-clover=coverage.clover
+
+      - name: Code coverage
+        if: "${{ matrix.php < '8.0' }}"
+        run: |
+          wget https://scrutinizer-ci.com/ocular.phar
+          php ocular.phar code-coverage:upload --format=php-clover coverage.clover --repository=g/minkphp/Mink
```
stof, line 64: I suggest removing this upload of the code coverage to scrutinizer. We don't actually follow it there, and for the static analysis, psalm and phpstan have become better alternatives IMO, so I'm considering disabling scrutinizer.

Note that in that case, we should of course stop collecting the code covering during tests too

.github/workflows/tests.yml
```diff
@@ -0,0 +1,64 @@
...
+          tools: composer
+
+      - name: Configure for PHP >= 7.1
+        if: "${{ matrix.php >= '7.1' }}"
+        run: |
+          composer require --no-update --dev symfony/error-handler "^4.4 || ^5.0"
+
+      - name: Install dependencies
+        run: |
+          composer install
```
stof, line 54: As we don't commit a lock file, using `composer update` would be better to actually describe what happens (a `composer install` without a lock file is actually running an update)

.github/workflows/tests.yml
```diff
@@ -0,0 +1,64 @@
...
+          - php: '5.4'
+          - php: '5.5'
+          - php: '5.6'
+          - php: '7.0'
+          - php: '7.1'
+          - php: '7.2'
+          - php: '7.3'
+          - php: '7.4'
+          - php: '8.0'
+          - php: '8.1'
```
stof, line 29: The old CI setup also had a job running tests with the min stability of composer set to `dev`

.github/workflows/tests.yml
```diff
@@ -0,0 +1,64 @@
+name: PHPUnit
```
stof, line 1: I would name it `CI` rather than `PHPUnit`, as we might add more jobs doing something else than running PHPUnit tests.

.github/workflows/tests.yml
```diff
@@ -0,0 +1,64 @@
...
+      - name: Checkout
+        uses: actions/checkout@v2
+        with:
+          fetch-depth: 1
+
+      - name: Setup PHP
+        uses: shivammathur/setup-php@v2
+        with:
+          coverage: "xdebug"
+          ini-values: date.timezone=Europe/Paris,memory_limit=-1,default_socket_timeout=10,session.gc_probability=0,apc.enable_cli=1,zend.assertions=1
```
stof, line 42: why all theses ? Aren't the default values of setup-php fine ?

.github/workflows/tests.yml
```diff
@@ -0,0 +1,64 @@
...
+          php-version: "${{ matrix.php }}"
+          extensions: "${{ env.extensions }}"
```
stof, line 44: there is no such env in the config, so this line looks useless

.github/workflows/tests.yml
alexpott: Removed as requested

.github/workflows/tests.yml
alexpott: Changed as requested

.github/workflows/tests.yml
alexpott: Added that for 7.4 and 8.0

.github/workflows/tests.yml
alexpott: Copied from Symfony - works without them so I've removed it.

stof (commented):
Please squash commits to have a clean history as well.

alexpott:
Latest run is all green... https://github.com/alexpott/Mink/actions/runs/1311054105

.github/workflows/tests.yml
```diff
@@ -0,0 +1,63 @@
...
+      matrix:
+        php: [ '5.4', '5.5', '5.6', '7.0', '7.1', '7.2', '7.3', '7.4', '8.0', '8.1']
+        composer_flags: [ '' ]
+        minimum_stability: [ '' ]
+        symfony_deprecations_helper: [ '' ]
+        include:
+          - php: 7.1
+            composer_flags: '--prefer-lowest --prefer-stable'
+            symfony_deprecations_helper: weak
+          - php: 8.0
```
stof, line 27: this needs to be quoted here too

stof (approved):

aik099:
@alexpott , please restore, in another PR, test suite coverage collection reporting, that was removed in other associated PR's as well.

// cc: @stof .

PR #813, opened 2021-10-10
Reinstate code coverage

alexpott:
@aik099 requested this. @stof asked me to remove it from the original MR but thinking about it we should only remove it in the MR the replaces it with psalm / phpstan


alexpott:
Sure but scutinizer is doing static analysis and presenting the code coverage results. So any MR that removes scutinizer should needs to cope with both I guess.

Chris53897:
This PR could be closed in my eyes. As Codecov is actual in use

PR #709, opened 2016-07-01
Element attribute not exists

J7mbo:

PR #817, opened 2021-12-07
Add element attribute not exists

murilohpucci:
Creating a pull request, if you have any questions, please let me know.

Thanks!

### DIFF
--- a/src/WebAssert.php
+++ b/src/WebAssert.php
@@ -566,6 +566,32 @@ class WebAssert
     }
 
     /**
+     * Checks that an attribute does not exist in an element.
+     *
+     * @param string       $selectorType
+     * @param string|array $selector
+     * @param string       $attribute
+     *
+     * @return NodeElement
+     *
+     * @throws ElementHtmlException
+     */
+    public function elementAttributeNotExists($selectorType, $selector, $attribute)
+    {
+        $element = $this->elementExists($selectorType, $selector);
+
+        $message = sprintf(
+            'The attribute "%s" was found in the %s.',
+            $attribute,
+            $this->getMatchingElementRepresentation($selectorType, $selector)
+        );
+
+        $this->assertElement(!$element->hasAttribute($attribute), $message, $element);
+
+        return $element;
+    }
+
+    /**
      * Checks that an attribute of a specific elements contains text.
      *
      * @param string       $selectorType
--- a/tests/WebAssertTest.php
+++ b/tests/WebAssertTest.php
@@ -1087,6 +1087,54 @@ class WebAssertTest extends TestCase
         );
     }
 
+    public function testElementAttributeNotExists()
+    {
+        $page = $this->getMockBuilder('Behat\\Mink\\Element\\DocumentElement')
+            ->disableOriginalConstructor()
+            ->getMock()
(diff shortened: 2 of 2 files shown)


murilohpucci:
Created the PR #817, I hope you can check. Thanks!

murilohpucci:
Squashed the commits into one to follow the CONTRIBUTING.md specs.

tests/WebAssertTest.php
```diff
@@ -1087,6 +1087,81 @@ public function testElementAttributeExists()
...
+        ;
+
+        $this->assertWrongAssertion(
+          'elementAttributeNotExists',
+          ['css', 'h2 > span', 'blah'],
+          'Behat\\Mink\\Exception\\ElementHtmlException',
+          'The attribute "blah" was found in the element matching css "h2 > span".'
+        );
+    }
+
```
aik099, line 1164: It's better to rewrite tests for an `elementAttributeNotExists` method based on tests for an `elementAttributeExists` rather than copy/pasting them from an #709.

aik099 (changes requested):

tests/WebAssertTest.php
murilohpucci: Hey! This is a good idea, I updated the code with a test based on the `elementAttributeExists`, I'm sorry for copy and pasting the code from #709, my idea with that was to fix the code review from the PR.

Thanks!

src/WebAssert.php
```diff
@@ -565,6 +565,32 @@ public function elementAttributeExists($selectorType, $selector, $attribute)
...
+    /**
+     * Checks that an attribute does not exist in an element.
+     *
+     * @param string       $selectorType
+     * @param string|array $selector
+     * @param string       $attribute
+     *
+     * @return NodeElement
+     *
+     * @throws ElementHtmlException|ElementNotFoundException
```
aik099, line 577: Please remove `|ElementNotFoundException` that you've added (when comparing with test for an `elementAttributeExists` method).

I've noticed, that none of the assertion methods using `$this->elementExists(...)` are telling that they're throwing an `ElementNotFoundException`.

murilohpucci: Oh, my bad, removed!

murilohpucci: Don't you think we should map this exception in tests? I mean, not now because this is out of the scope of this issue, but in another issue. Since the function can throw this exception.

aik099: > Don't you think we should map this exception in tests? I mean, not now because this is out of the scope of this issue, but in another issue. Since the function can throw this exception.

No. Testing of assertion method IMO should test both passed and failed assertion and nothing more. It's out of the scope of assertion method testing to verify the internal logic of methods used inside it (e.g. for finding an element on the page). Also, it's likely not documented (that exception would be thrown on a missing element).

aik099 (approved):
Ready to merge.

src/WebAssert.php
murilohpucci: Thanks a lot for this response, it makes sense!

murilohpucci:
That's awesome @aik099, thanks!

stof (approved):

PR #826, opened 2022-04-14
Add support for executeScript/evaluateScript args

mvorisek:
The args support is necessary as webdriver objects must be passed natively to the underlaying drivers & cannot be serialized directly into the 1st script argument.

Example `executeScript` method prototype of one of the most popular driver - https://github.com/php-webdriver/php-webdriver/blob/8ffa927b270e932449e8015abf4d38bb0eff24b7/lib/Remote/RemoteWebDriver.php#L324

### DIFF
--- a/src/Driver/CoreDriver.php
+++ b/src/Driver/CoreDriver.php
@@ -428,7 +428,7 @@ abstract class CoreDriver implements DriverInterface
     /**
      * {@inheritdoc}
      */
-    public function executeScript($script)
+    public function executeScript($script, array $args = [])
     {
         throw new UnsupportedDriverActionException('JS is not supported by %s', $this);
     }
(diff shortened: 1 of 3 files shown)


aik099:
Adding more parameters to the `DriverInterface` interface might be a BC break (not sure if an optional argument causes BC breaks). I guess you'll need to create PR for every driver to pass through these arguments. I guess the Selenium2 is the only one capable of it.

mvorisek:
Yes, it is fatal error BC break - https://3v4l.org/kO2hW.

I will create PRs to these drivers - https://github.com/minkphp/MinkSelenium2Driver and https://github.com/silverstripe/MinkFacebookWebDriver shortly. What are the other popular drivers?

stof:
We simply cannot do such BC break in a minor version. That would mean we stop applying semver.

mvorisek:
Are there any objections to release a v2?

PR #827, opened 2022-04-18
Optimize single element find

mvorisek:
no BC, no functionality change

improve performance with large sites with wide locators


mvorisek:
please help with adjusting the tests

src/Element/Element.php
```diff
@@ -165,9 +158,31 @@ public function findAll($selector, $locator)
         $xpath = $this->selectorsHandler->selectorToXpath($selector, $locator);
         $xpath = $this->xpathManipulator->prepend($xpath, $this->getXpath());
 
+        if ($firstOnly) {
+            $xpath = '(' . $xpath . ')[1]';
+        }
+
```
aik099, line 164: I think, but not 100% certain, that this XPath altering code is better to be paced in classes, that builds these XPaths (e.g. the `\Behat\Mink\Selector\SelectorInterface::translateToXPath` methods).

stof: @aik099 those classes don't know whether you call `find` or `findAll`

stof: thus, they run before the prepending of the current element, so they would produce the wrong result.

mvorisek: yes, private method to dedup the code

mvorisek:
> @mvorisek you need to update the tests of `find()`

Yes, I tried to do so, but the assertions are written using mocked objects and I do not know how to fix them. Can you please help?

PR #831, opened 2022-05-04
Deprecate keyPress, keyDown, keyUp in favor of pressKey.

mparker17:
## Problem / motivation

in #828, @stof said that "the [driver methods] that should actually be deprecated are the separate keydown, keypress and keyup methods in favor of a single pressKey method to match what is actually possible in browsers."

## Proposed resolution

* Deprecate `DriverInterface::keyDown()`, `DriverInterface::keyPress()`, and `DriverInterface::keyUp()` in favor of a new `DriverInterface::pressKey()`
* Deprecate `NodeElement::keyDown()`, `NodeElement::keyPress()`, and `NodeElement::keyUp()` in favor of a new `NodeElement::pressKey()`

## Notes

I modelled the `DriverInterface::pressKey()` method after [php-webdriver/php-webdriver's `\Facebook\WebDriver\Remote\RemoteKeyboard::sendKeys()`](https://github.com/php-webdriver/php-webdriver/blob/b27ddf458d273c7d4602106fcaf978aa0b7fe15a/lib/Remote/RemoteKeyboard.php#L31-L48)

### DIFF
--- a/src/Element/NodeElement.php
+++ b/src/Element/NodeElement.php
@@ -306,25 +306,38 @@ class NodeElement extends TraversableElement
     }
 
     /**
+     * Press one or more keyboard keys on an element.
+     *
+     * @param string|int|array $keys a string to type ('PHP Hypertext Preprocessor'), a integer key code (88), or an array of keys to press in sequence (e.g.: ["\u{E008}", 'H', "\u{E001}", 'ello']). See https://w3c.github.io/webdriver/#keyboard-actions
+     */
+    public function pressKey($keys) {
+        $this->getDriver()->pressKey($this->getXpath(), $keys);
+    }
+
+    /**
      * Presses specific keyboard key.
      *
      * @param string|int  $char     could be either char ('b') or char-code (98)
      * @param string|null $modifier keyboard modifier (could be 'ctrl', 'alt', 'shift' or 'meta')
+     *
+     * @deprecated use pressKey($keys) instead.
      */
     public function keyPress($char, $modifier = null)
     {
-        $this->getDriver()->keyPress($this->getXpath(), $char, $modifier);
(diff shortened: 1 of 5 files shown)


src/Element/NodeElement.php
```diff
@@ -305,37 +305,52 @@ public function blur()
         $this->getDriver()->blur($this->getXpath());
     }
 
+    /**
+     * Press one or more keyboard keys on an element.
+     *
+     * @param string|int|array $keys a string to type ('PHP Hypertext Preprocessor'), a integer key code (88), or an array of keys to press in sequence (e.g.: ["\u{E008}", 'H', "\u{E001}", 'ello']). See https://w3c.github.io/webdriver/#keyboard-actions
+     */
+    public function pressKey($keys) {
```
aik099, line 313: Changing an interface is a BC break.

src/Element/NodeElement.php
```diff
@@ -305,37 +305,52 @@ public function blur()
...
      *
      * @param string|int  $char     could be either char ('b') or char-code (98)
      * @param string|null $modifier keyboard modifier (could be 'ctrl', 'alt', 'shift' or 'meta')
+     *
+     * @deprecated use pressKey($keys) instead.
      */
     public function keyPress($char, $modifier = null)
     {
-        $this->getDriver()->keyPress($this->getXpath(), $char, $modifier);
+        $this->pressKey($this->wrapCharWithModifier($char, $modifier));
```
aik099, line 327: Adding a `@deprecated` tag alone in the DocBlock won't help much, but having a matching deprecation notice triggering code will.

---

The same goes for other similar places in this class.

aik099 (changes requested):

PR #838, opened 2022-10-29
fix: avoid a TypeError if findAll returns a null

svemir:
On PHP 8.1, when running some internal tests I was getting a TypeError because the driver would return a null for findAll.

This adds an is_array check before calling count, just in case.

### DIFF
--- a/src/Element/Element.php
+++ b/src/Element/Element.php
@@ -145,7 +145,7 @@ abstract class Element implements ElementInterface
     {
         $items = $this->findAll($selector, $locator);
 
-        return count($items) ? current($items) : null;
+        return is_array($items) && count($items) ? current($items) : null;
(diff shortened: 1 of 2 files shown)


svemir:
Perhaps all drivers are guaranteed to always return an array from findAll? The interface suggests that, but I am not sure if it is being actually enforced for all of them. My use case was somewhat convoluted, so feel free to just close this if it can't happen in real world.

aik099 (approved):

stof:
If a driver does not return an array, it does not respect the contract of the interface. And that must be fixed in the driver instead.
All official drivers are properly returning arrays.

So to me, this is a -1

aik099:
@svemir , what driver are you using?

PR #841, opened 2023-05-12
Feature/allow phpunit 10

Chris53897:
The 2 abstract Test-Classes are excludet to nor raise Warnings and break CI-runs.

### DIFF
--- a/phpunit.xml.dist
+++ b/phpunit.xml.dist
@@ -9,10 +9,7 @@
   <testsuites>
     <testsuite name="Behat Mink test suite">
       <directory>tests</directory>
-      <exclude>./tests/Element/ElementTest.php</exclude>
-      <exclude>./tests/Selector/NamedSelectorTest.php</exclude>
     </testsuite>
-
   </testsuites>
 
   <source>
@@ -20,5 +17,12 @@
       <directory>./src</directory>
     </include>
   </source>
+
+  <!--
+  Not supported in PHPUnit
+  <extensions>
+    <bootstrap class="Symfony\Bridge\PhpUnit\SymfonyTestsListener" />
+  </extensions>
+  -->
 
 </phpunit>
--- a/phpunit.xml.dist
+++ b/phpunit.xml.dist
@@ -9,7 +9,10 @@
   <testsuites>
     <testsuite name="Behat Mink test suite">
       <directory>tests</directory>
+      <exclude>./tests/Element/ElementTest.php</exclude>
+      <exclude>./tests/Selector/NamedSelectorTest.php</exclude>
     </testsuite>
+
   </testsuites>
 
   <source>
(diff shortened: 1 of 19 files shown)


phpunit.xml.dist
```diff
@@ -1,19 +1,24 @@
...
+  </testsuites>
+
+  <source>
+    <include>
+      <directory>./src</directory>
+    </include>
+  </source>
 
-    <listeners>
-        <listener class="Symfony\Bridge\PhpUnit\SymfonyTestsListener" />
```
stof, line 17: This makes us loose the existing feature where our CI fails if we trigger deprecated APIs, which is something we wanted to have

phpunit.xml.dist
```diff
@@ -1,19 +1,24 @@
...
-    <testsuites>
-        <testsuite name="Behat Mink test suite">
-            <directory>tests</directory>
-        </testsuite>
-    </testsuites>
+  <testsuites>
+    <testsuite name="Behat Mink test suite">
+      <directory>tests</directory>
+      <exclude>./tests/Element/ElementTest.php</exclude>
+      <exclude>./tests/Selector/NamedSelectorTest.php</exclude>
```
stof, line 13: excluding some tests is a no-go. Our testsuite must run.

If those are not actually tests but abstract test cases, the right fix is to rename them rather than having to maintain exclude rules.

stof (changes requested):

phpunit.xml.dist
Chris53897: Yes of course. This --migrate-configuration does remove it, and i did not checked the backup file. I will re-add this

stof: Actually, the symfony/phpunit-bridge is not yet compatible with PHPUnit 10. So this is a blocker for using it in our CI

Chris53897: I just noticed that ;(
I added it, but outcommented the code. Xml-Names changed too, i am not sure if phunit 8 can handle the new Xml-Names.

phpunit.xml.dist
Chris53897: These are abstract Tests. I changed the tests. There are of course multiple solutions. Please check if my solution is ok.

phpunit.xml.dist
Chris53897: I guess the best solution is, to wait until the bridge symfony/phpunit-bridge is compatible with PHPUnit 10.
There could be 2 seperate phpunit (for 8/9 and 10) config files and a switch in the ci tests.
But i guess that is not what we want.

stof: Indeed. There is nothing wrong with using PHPUnit 9 in our CI. We don't want to maintain 2 configurations if we can avoid it.

However, the TestCase renaming and the migration to static methods for data-providers are something we could merge even if we don't use PHPUnit 10 yet.

Chris53897: Ok, i will provide a seperate PR. (probably this weekend)

aik099:
This PR also includes the changes from #842 . Not sure if that was planned.

Chris53897:
@aik099 Yes this in intended, please see https://github.com/minkphp/Mink/pull/841#discussion_r1192433462
This PR could be rebased after the bridge symfony/phpunit-bridge is compatible with PHPUnit 10.

PR #842, opened 2023-05-14
tests: prepare for phpunit 10

Chris53897:
add static data generators
fix deprecated file-structure for abstract test classes

### DIFF
--- a/tests/Driver/CoreDriverTest.php
+++ b/tests/Driver/CoreDriverTest.php
@@ -74,7 +74,7 @@ class CoreDriverTest extends TestCase
         $coreDriverMethod->invokeArgs($driver, $this->getArguments($method));
     }
 
-    public function getDriverInterfaceMethods()
+    public static function getDriverInterfaceMethods()
     {
         $ref = new \ReflectionClass('Behat\Mink\Driver\DriverInterface');
 
--- a/tests/Element/DocumentElementTest.php
+++ b/tests/Element/DocumentElementTest.php
@@ -5,7 +5,7 @@ namespace Behat\Mink\Tests\Element;
 use Behat\Mink\Element\DocumentElement;
 use Behat\Mink\Element\NodeElement;
 
-class DocumentElementTest extends ElementTest
+class DocumentElementTest extends ElementTestCase
 {
     /**
      * Page.
--- a/tests/Element/ElementTestCase.php
+++ b/tests/Element/ElementTestCase.php
@@ -8,7 +8,7 @@ use Behat\Mink\Session;
 use PHPUnit\Framework\MockObject\MockObject;
 use PHPUnit\Framework\TestCase;
 
-abstract class ElementTest extends TestCase
+abstract class ElementTestCase extends TestCase
 {
     /**
      * Session.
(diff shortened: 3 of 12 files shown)


tests/Element/AbstractElement.php
```diff
@@ -8,7 +8,7 @@
 use PHPUnit\Framework\MockObject\MockObject;
 use PHPUnit\Framework\TestCase;
 
-abstract class ElementTest extends TestCase
+abstract class AbstractElement extends TestCase
```
aik099, line 11: What was the purpose of this class renaming?

tests/Selector/AbstractNamedSelector.php
```diff
@@ -6,7 +6,7 @@
 use Behat\Mink\Selector\Xpath\Escaper;
 use PHPUnit\Framework\TestCase;
 
-abstract class NamedSelectorTest extends TestCase
+abstract class AbstractNamedSelector extends TestCase
```
aik099, line 9: What was the purpose of this class renaming?

Chris53897: PHPUnit 10 will raise a warning, if there are classes with the Suffix `Test` and no tests or abstract Classes. https://github.com/sebastianbergmann/phpunit/issues/4979#issuecomment-1144781859
https://github.com/sebastianbergmann/phpunit/issues/4979#issuecomment-1461594789

tests/Element/AbstractElement.php
Chris53897: Same as below

tests/Selector/AbstractNamedSelector.php
Chris53897: The CI Tests failed because of this

stof: I would actually use `TestCase` as suffix here

Chris53897: ok, I removed the prefix and added your suggested suffix.

tests/Element/DocumentElementTestCaseTest.php
```diff
@@ -4,7 +4,7 @@
 
 use Behat\Mink\Element\DocumentElement;
 
-class DocumentElementTest extends ElementTest
+class DocumentElementTestCaseTest extends ElementTestCase
```
stof, line 7: This should still be named `DocumentElementTest`. They are the tests for `DocumentElement`

stof (changes requested):

aik099 (commented):
@Chris53897 , what is the general purpose of renaming `*Test` classes into `*TestCase` classes?

stof:
@aik099 PHPUnit 10 does not allow have abstract classes ending with `Test` (older versions of PHPUnit were silently considering that those were not actually tests that should be loaded in the testsuite)

Chris53897:
I changed the Filenames. I hope everything is correct now

tests/Driver/CoreDriverTest.php
```diff
@@ -22,9 +22,15 @@ public function testNoExtraMethods()
 
     public function testCreateNodeElements()
     {
+        if(\PHPUnit\Runner\Version::id() >= 10) {
```
stof: this should be removed from this PR as we are *not* using phpunit 10 in our setup (intentionally).

Chris53897: i removed it now

stof (approved):

PR #847, opened 2023-06-10
Improve PhpStorm's auto-complete of "DriverInterface" method arguments

uuf6429:
The changes are fully backword compatible.

### DIFF
--- a/src/Driver/DriverInterface.php
+++ b/src/Driver/DriverInterface.php
@@ -6,12 +6,6 @@
  *
  * For the full copyright and license information, please view the LICENSE
  * file that was distributed with this source code.
- *
- * @noinspection PhpLanguageLevelInspection
- * @noinspection PhpMissingParamTypeInspection
- * @noinspection PhpMissingReturnTypeInspection
- * @noinspection ReturnTypeCanBeDeclaredInspection
- * @noinspection PhpElementIsNotAvailableInCurrentPhpVersionInspection
  */
 
 namespace Behat\Mink\Driver;
@@ -21,7 +15,6 @@ use Behat\Mink\Exception\DriverException;
 use Behat\Mink\Exception\UnsupportedDriverActionException;
 use Behat\Mink\KeyModifier;
 use Behat\Mink\Session;
-use JetBrains\PhpStorm\ExpectedValues;
 use JetBrains\PhpStorm\Language;
 
 /**
@@ -167,8 +160,8 @@ interface DriverInterface
     /**
      * Sets HTTP Basic authentication parameters.
      *
-     * @param string|false $user user name or false to disable authentication
-     * @param string $password password
+     * @param string|false $user     user name or false to disable authentication
+     * @param string       $password password
      *
      * @return void
      *
@@ -229,7 +222,7 @@ interface DriverInterface
      *
      * Passing null as value will delete the cookie.
      *
-     * @param string $name
+     * @param string      $name
      * @param string|null $value
      *
      * @return void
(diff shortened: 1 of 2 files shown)


src/Driver/DriverInterface.php
```diff
@@ -1,11 +1,17 @@
...
  * (c) Konstantin Kudryashov <ever.zet@gmail.com>
  *
  * For the full copyright and license information, please view the LICENSE
  * file that was distributed with this source code.
+ *
+ * @noinspection PhpLanguageLevelInspection
+ * @noinspection PhpMissingParamTypeInspection
+ * @noinspection PhpMissingReturnTypeInspection
+ * @noinspection ReturnTypeCanBeDeclaredInspection
+ * @noinspection PhpElementIsNotAvailableInCurrentPhpVersionInspection
```
uuf6429, line 14: I've mixed feeling if this should be here: Disabling these inspections makes it nicer for the developer to focus on the bigger problems, but it's also basically hiding some problems.

src/Driver/DriverInterface.php
```diff
@@ -159,8 +167,8 @@ public function back();
     /**
      * Sets HTTP Basic authentication parameters.
      *
-     * @param string|false $user     user name or false to disable authentication
-     * @param string       $password password
```
uuf6429, line 163: Unfortunately, the IDE has this style rule not enabled by default.
If it's preferred to have this sort of style, I can roll it back.

src/Driver/DriverInterface.php
```diff
@@ -581,13 +658,19 @@ public function blur($xpath);
      * @throws UnsupportedDriverActionException When operation not supported by the driver
      * @throws DriverException                  When the operation cannot be done
      */
-    public function keyPress($xpath, $char, $modifier = null);
+    public function keyPress(
+        #[Language('XPath')]
+        $xpath,
+        $char,
+        #[ExpectedValues([null, 'ctrl', 'alt', 'shift', 'meta'])]
```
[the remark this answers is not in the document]
uuf6429: Ah, sorry I missed that. I'll roll it back. 👍

uuf6429:
@stof any chance getting this PR merged?

Sadly, I keep ending up adding such attributes at the driver level, which only benefits the drivers themselves (and would be redundant with this PR merged).

aik099 (approved):

aik099:
Merging, thank you @uuf6429 .

Please explain the difference in working with `DriverInterface` with/without having these PhpStorm-specific attributes available.

uuf6429:
> Please explain the difference in working with DriverInterface with/without having these PhpStorm-specific attributes available.

If you mean why those attributes are useful:

Participating IDEs (mainly JetBrains) should be able to highlight such parameters, which improves developer experience for maintainers but also direct driver users.
image
(instead of a plain green text for the entire strings)

At the moment, it works for me already, because the IDE realises that I'm using the classic driver, which already has those attributes.

For driver maintainers, unless they add the attributes themselves, they also will not get that hinting. These attributes are useful for driver maintainers as well, since drivers tend to call such methods themselves.

Hope that answers your question, let me know if you meant something else.

stof:
Projects should almost never use the driver directly though. They are expected to work with the high-level Mink API.

uuf6429:
@stof agreed, I believe there are other cases in mink where it makes sense to support that.

For example, in my current project, we developed our own selector/locator system with slightly better DX that is still based on mink:
image

(maybe at one point I'll make it into a mink PR)

----

In any case, I believe those attributes are low-effort/impact quick wins.

PR #863, opened 2023-12-17
Fixed argument type in switchToIFrame() to be integer.

AlexSkrypnyk:
Release `1.11.0` introduced strict types.

With that, the `switchToIFrame(string? $id)` method started to fail on consumer projects using `w3c` = `true` because the expected `$id` argument has a type `string` instead of expected `int`. 

According to https://www.w3.org/TR/webdriver/#switch-to-frame, the `$id` is expected to be an integer.

This PR updates `string?` with `int?`.

Note that downstream project `MinkSelenium2Driver` have the same issue propagated after the recent release `1.7.0`: https://github.com/minkphp/MinkSelenium2Driver/blob/master/src/Selenium2Driver.php#L431

### DIFF
--- a/src/Driver/CoreDriver.php
+++ b/src/Driver/CoreDriver.php
@@ -160,14 +160,14 @@ abstract class CoreDriver implements DriverInterface
     }
 
     /**
-     * @param string|null $name
+     * @param int|null $name
      *
      * @return void
      *
      * @throws UnsupportedDriverActionException When operation not supported by the driver
      * @throws DriverException                  When the operation cannot be done
      */
-    public function switchToIFrame(?string $name = null)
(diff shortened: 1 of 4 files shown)


aik099:
@AlexSkrypnyk , please see #866 for a complete story.

The problem is that until the https://github.com/minkphp/MinkSelenium2Driver/pull/382 the attempt to operate on frames by specifying their name in w3c mode wasn't working.

You have several options now:

1. either wait for https://github.com/minkphp/MinkSelenium2Driver release with a fix (the right clicking problem still not fixed)
2. or switch to the https://github.com/minkphp/webdriver-classic-driver , that also supports Selenium 4

Issue #779, opened 2019-06-17
Selectors should recognise aira-label and aria-labelled-by where appropriate


timhunt:
The selectors defined in https://github.com/minkphp/Mink/blob/master/src/Selector/NamedSelector.php do a good job of recognising most HTML. E.g. they will find a checkbox by the associated <label for="...">, or a link but the alt text of the icon <img> inside the <a>.

However, at present they do not find e.g. <input type="checkbox" aria-label="Frog">, or using aria-labelled-by="id" on the checkbox, rather than for="id" on the label. It would be nice to fix this.

(This is something we would like in the Moodle project. See https://tracker.moodle.org/browse/MDL-65940.)

bkosborne:
Created a PR that adds this support for buttons at least.

aik099:
@timhunt , does the https://github.com/minkphp/Mink/pull/865 PR fully implement the functionality you're talking about?

@bkosborne , the mentioned PR adds support for [aria-label](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Attributes/aria-label) attribute, but not the [aria-labelledby](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Attributes/aria-labelledby) attribute. Are you planning to add another PR for that (or adjust existing PR)?

PR #865, opened 2024-02-01
Add support for finding buttons that use aria-label for a value.

bkosborne:
See #779 

This adds support for finding a button by its aria-label value.


maelanleborgne:
As it is very hard to extend the library, I would greatly appreciate if this feature was merged :slightly_smiling_face:

aik099 (approved):

aik099:
> As it is very hard to extend the library, I would greatly appreciate if this feature was merged 🙂

@maelanleborgne , can you please be more specific?

P.S.
I've approved this PR.

I've confirmed, that the tests actually test the added code by removing added HTML elements from the test fixtures one by one and witnessing, that tests start to fail after that.

// cc: @stof

uuf6429:
Not bad, but `aria-label` applies to way more elements than just `<button>`. In particular, it should probably be implemented on `'%linkMatch%'` (similar to `'%buttonMatch%'`).

